PR #34, opened 2026-04-05
build the "prebuilt" libraries from sources

mwieder:
OK - this is a big pull request but there are a lot of moving parts.
The PR removes the need to build the prebuilt archives separately and downloads the source files when necessary. From a cold start the compilation process takes more time, but it's more than offset by the elimination of the need to create and package the archives, and it makes updating the libraries painless.
The original idea for this is based on some unmerged pull prequests on the original livecode archive.


mwieder:
closing for now... git problem.

PR #38, opened 2026-04-08
build libraries from source

mwieder:
OK - this is a big pull request but there are a lot of moving parts.
The PR removes the need to build the prebuilt archives separately and downloads the source files when necessary. From a cold start the compilation process takes more time, but it's more than offset by the elimination of the need to create and package the archives, and it makes updating the libraries painless.
The original idea for this is based on some unmerged pull requests on the original livecode archive.


mwieder:
Oh, yes. The thirdparty prebuilt zip archives are no longer necessary. From a cold start the required sources are downloaded from their web locations. Updating one of the libraries is now mostly a matter of changing the prebuilt/versions/<libraryname> file. Only the first line in the version file is significant, so adding a new first line allows you to keep the previous version for reference.

mwieder:
I'm on linux, so it works for me here. I haven't been able to compile on osx because xcode keeps throwing stupid linker errors. There shouldn't be a problem otherwise. And in theory this should work on Windows.

mwieder:
I'll see if I can dig it up... something about a bad argument being passed to ld. I know how to pass arguments to ld explicitly, but as far as I can tell there's nothing that's doing that.

mwieder:
Here's the errant code - the invocation and the error message. I see the "-Xlinker -no_adhoc_codesign" linker option in the calling arguments, but I can't find where that comes in. Grep doesn't show anything and I haven't found it in any xcode config files. Googling on it doesn't help either. Any help appreciated.

Ld /Users/pogothepossum/Documents/livecode/_build/mac/Debug/perfect-target normal (in target 'perfect' from project 'perfect')

    cd /Users/pogothepossum/Documents/livecode/util/perfect

    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -Xlinker -reproducible -target x86_64-apple-macos14.2 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk -O0 -L/Users/pogothepossum/Documents/livecode/_cache/mac/EagerLinkingTBDs/Debug -L/Users/pogothepossum/Documents/livecode/_build/mac/Debug -F/Users/pogothepossum/Documents/livecode/_cache/mac/EagerLinkingTBDs/Debug -F/Users/pogothepossum/Documents/livecode/_build/mac/Debug -filelist /Users/pogothepossum/Documents/livecode/_cache/mac/Debug/perfect.build/Objects-normal/x86_64/perfect-target.LinkFileList -Xlinker -object_path_lto -Xlinker /Users/pogothepossum/Documents/livecode/_cache/mac/Debug/perfect.build/Objects-normal/x86_64/perfect-target_lto.o -Xlinker -no_deduplicate -Wl,-platform_version -Wl,macos -Wl,14.2 -Xlinker -no_adhoc_codesign -Xlinker -dependency_info -Xlinker /Users/pogothepossum/Documents/livecode/_cache/mac/Debug/perfect.build/Objects-normal/x86_64/perfect-target_dependency_info.dat -o /Users/pogothepossum/Documents/livecode/_build/mac/Debug/perfect-target

ld: -platform_version: malformed version number '-no_adhoc_codesign' cannot fit in 32-bit xxxx.yy.zz

clang: error: linker command failed with exit code 1 (use -v to see invocation)

emily-elizabeth:
This is what Claude says about the issue:

The error isn't caused by -no_adhoc_codesign being in the wrong place — that flag is fine. The real problem is that -platform_version is missing its third argument.

-platform_version requires three arguments: platform, min_version, sdk_version. In the linker command you can see:

-Wl,-platform_version -Wl,macos -Wl,14.2

That's only two arguments (macos and 14.2). The linker then greedily consumes the next token — -no_adhoc_codesign — as the third argument (the SDK version), and then fails trying to parse it as a version number.

-no_adhoc_codesign itself is injected automatically by Xcode 15+'s build system into every link invocation — it doesn't come from your source, your GYP files, or any Xcode config you'd write. That's why grep finds nothing.

The fix is wherever -platform_version,macos,14.2 is being generated, a third version argument needs to be added — e.g. -Wl,-platform_version -Wl,macos -Wl,14.2 -Wl,14.2. That's most likely coming from an OTHER_LDFLAGS entry or a deployment target setting in either common.gypi or the perfect target's GYP configuration.

PR #53, opened 2026-04-10
Migrate engine source to OpenSSL 3.x (closes #50)

### LINKED ISSUE #50: prebuilt/fetch-libraries.sh: fetchLibrary call passes '${ARCH}URL' literal instead of per-library URL (breaks every build since ba099e69)
Hi! I hit a build failure on `main` that traces back to a quoting bug in `prebuilt/fetch-libraries.sh`. Wanted to flag it and propose a fix. It's a one-line root cause, but I've tried to write the fix out as copy-pasteable terminal commands so you don't have to fight with git.

## Symptom

On a clean `make compile-mac` from `main`, the `fetch-mac` Xcode phase fails with:

```
note: Fetching prebuilt libraries for OSX
Fetching remote library: UniversalURL/OpenSSL-1.1.1v-mac-Universal-PIC
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (6) Could not resolve host: UniversalURL
Failed to find library OpenSSL-1.1.1v-mac-Universal-PIC either remotely or locally
Command PhaseScriptExecution failed with a nonzero exit code
** BUILD FAILED **
```

`UniversalURL` isn't a typo or a placeholder — it's what the shell script actually assembles at runtime. The same bug hits every other platform too, just with different nonsense hostnames (`x86_64URL`, `armv7URL`, `jsURL`, …).

## Root cause

In `prebuilt/fetch-libraries.sh`, the two `fetchLibrary` call sites inside the platform loop pass `"${ARCH}URL"` as the first argument:

```bash
for ARCH in ${SELECTED_ARCHS} ; do
    for LIB in "${LIBS[@]}" ; do
        if [ ! -z "${SUBPLATFORMS}" ] ; then
            for SUBPLATFORM in "${SUBPLATFORMS[@]}" ; do
                fetchLibrary "${ARCH}URL" "${LIB}" "${PLATFORM}" "${ARCH}" "${SUBPLATFORM}"
            done
        else
            fetchLibrary "${ARCH}URL" "${LIB}" "${PLATFORM}" "${ARCH}"
        fi
    done
done
```

`"${ARCH}URL"` is direct string concatenation, not variable indirection. `${ARCH}` expands to `Universal` / `x86_64` / `armv7` / etc., and `URL` stays literal. So the first arg is the string `UniversalURL` — and inside `fetchLibrary` it's used verbatim as a URL prefix:

```bash
function fetchLibrary {
    local LIBURL=$1
    ...
    curl -k "${LIBURL}/${NAME}" -o "${FETCH_DIR}/${NAME}.tar.bz2" --fail
}
```

So curl ends up trying to fetch `UniversalURL/OpenSSL-1.1.1v-mac-Universal-PIC` and treats `UniversalURL` as a hostname → `Could not resolve host`.

The intent was clearly "look up a per-library URL," since the file defines these at the top:

```bash
URLCURL="https://github.com/curl/curl/archive/refs/heads/master.zip"
URLOPENSSL="https://github.com/openssl/openssl/archive/refs/heads/master.zip"
URLICU="https://github.com/unicode-org/icu/archive/refs/heads/main.zip"
URLCEF="https://github.com/chromiumembedded/cef/archive/refs/heads/master.zip"
URLSQLITE="https://sqlite.org/2023/sqlite-amalgamation-3420000.zip"
```

Those are keyed by **library name** (`URL<LIB>`), not by architecture. The call site should be reading `URL${LIB}` and then indirect-expanding it (bash `${!name}` syntax) to get the actual URL.

## When it landed

Commit ba099e69 ("Updated gyp files / Updated build files rather than fetch", 2026-03-28) reworked `fetch-libraries.sh` to take a per-library URL but miswrote the call-site indirection. The same commit also deleted a helpful local-cache guard that used to short-circuit the fetch when `prebuilt/lib/<platform>/` was already populated — so on a tree where someone has built the libs manually (which is what BUILD1.md tells you to do), the broken fetch *still* runs and blows up the build.

Only one commit has touched `fetch-libraries.sh` since then — 318558ce ("finally compiles/runs") — and it was just a whitespace tweak to the linux ARCH list, so the bug has been sitting in `main` ever since.

## Fix — copy-pasteable terminal commands

Open Terminal, `cd` into the repo, and run these four commands. Each does exactly one thing so you can stop after any of them if something looks wrong.

```bash
cd ~/path/to/HyperXTalk      # adjust to wherever your clone lives
```

**1. Fix the two call sites** — change `"${ARCH}URL"` → `"URL${LIB}"` so the lookup is keyed by library name:

```bash
sed -i '' 's|fetchLibrary "${ARCH}URL"|fetchLibrary "URL${LIB}"|g' prebuilt/fetch-libraries.sh
```

**2. Fix the `LIBURL` assignment** inside `fetchLibrary` — use bash indirect expansion so `$1` is interpreted as a variable name, not a literal string:

```bash
sed -i '' 's|local LIBURL=$1|local LIBURL="${!1}"|' prebuilt/fetch-libraries.sh
```

**3. Verify** the three lines look right:

```bash
grep -n 'fetchLibrary "URL\|local LIBURL' prebuilt/fetch-libraries.sh
```

You should see something like:

```
 57:	local LIBURL="${!1}"
183:					fetchLibrary "URL${LIB}" "${LIB}" "${PLATFORM}" "${ARCH}" "${SUBPLATFORM}"
186:				fetchLibrary "URL${LIB}" "${LIB}" "${PLATFORM}" "${ARCH}"
```

**4. Commit** (if you want to — standard flow):

```bash
git add prebuilt/fetch-libraries.sh
git commit -m "fetch-libraries.sh: fix ARCH/LIB mix-up in fetchLibrary call site

The first arg passed to fetchLibrary was '\${ARCH}URL' which is direct
string concatenation, producing literal strings like 'UniversalURL' that
get fed to curl as hostnames. The intent was indirect variable lookup
keyed by library name. Switch the call site to 'URL\${LIB}' and use bash
indirect expansion (\${!1}) inside fetchLibrary to dereference it to
\$URLOPENSSL, \$URLICU, etc.

Introduced in ba099e69."
```

> A note on `sed -i ''`: that empty `''` after `-i` is the macOS/BSD version. If you ever run these on Linux, drop the `''` — GNU sed takes plain `-i`.

## Caveat / further work

Wiring the call-site indirection up correctly will stop the build from failing with "Could not resolve host," but it won't make the fetch itself succeed. The `$URL<LIB>` values in the file currently point at GitHub `archive/refs/heads/master.zip` URLs, which are source tarballs — not prebuilt `libOpenSSL-<version>-<platform>-<arch>-PIC.tar.bz2` archives, which is what `fetchLibrary` expects by the constructed name. The old LiveCode prebuilts CDN (`https://downloads.livecode.com/prebuilts`, which `ba099e69` commented out) seems to be gone. So after the fix, you'll still need to either:

- point the `$URL<LIB>` variables at somewhere that actually hosts the LiveCode-style prebuilt tarballs, or
- route `fetch-mac` (in `prebuilt/fetch.gyp`) through `build-libraries.sh` instead of `fetch-libraries.sh` — which looks like it was the original intent behind ba099e69 given its commit title "Updated build files rather than fetch". `build-libraries.sh` compiles each lib from source via `prebuilt/scripts/build-*.sh` and populates `prebuilt/lib/<platform>/` directly, no network fetch required.

The second option is closer to what most contributors will want — it's self-contained and reproducible. Happy to open a follow-up PR for that if it would help; just let me know which direction you'd prefer.

## Optional: restore the cache guard

If you'd like the "skip fetch when libs are already present" short-circuit back (it was in the original version of the file before ba099e69), paste this block just after the `SCRIPT_DIR=$(cd ... )` line near the top of the file:

```bash
# Skip fetch if the output library directory for the requested platform
# is already populated (e.g. libraries were built from source or extracted
# manually). Avoids hitting the network on local builds.
if [ "$#" -ge 1 ]; then
    _FETCH_PLATFORM=$(basename "$1")
    _LIB_OUT="${SCRIPT_DIR}/lib/${_FETCH_PLATFORM}"
    if [ -d "${_LIB_OUT}" ] && [ -n "$(ls -A "${_LIB_OUT}" 2>/dev/null)" ]; then
        echo "note: prebuilt libraries for ${_FETCH_PLATFORM} already present at ${_LIB_OUT}, skipping fetch"
        exit 0
    fi
fi
```

That way contributors who've manually populated `prebuilt/lib/mac/` via BUILD1.md can build cleanly without ever touching the fetch path.

---

Thanks for maintaining this fork! Let me know if any of the above is unclear or if you'd prefer I just open a PR with the fix.

docmeth02:
Follow-up to #52. This is the actual OpenSSL 1.1.1 → 3.x engine migration for issue #50 — #52 only restored Mac builds using a 1.x compat shim in `build-mac-extras.sh`, this PR rips out the shim and moves the engine code to the 3.x API so libcurl ≥ 8 (which needs OpenSSL ≥ 3) can be next.

**Marking as draft — would love a code review + linux verification before we finalize and merge.**

## What's in it

- Rename the four symbols that actually got renamed in 3.0:
  `EVP_CIPHER_CTX_block_size`/`_key_length`, `EVP_CIPHER_key_length`, `SSL_get_peer_certificate` → `*_get_*` / `SSL_get1_peer_certificate`. These are the only symbols removed from the 3.x ABI; everything else `ssl.stubs` lists is header-deprecated but still exported.
- Drop the libcustomcrypto/libcustomssl compat shim from `build-mac-extras.sh`. No more injected `.o` wrappers.
- Load the `legacy` provider in `InitSSLCrypt()` (guarded by `OPENSSL_VERSION_NUMBER >= 0x30000000L`) so `encrypt`/`decrypt` with `rc4`, `des`, `blowfish`, etc. keeps working. OpenSSL 3 moved these into a provider that isn't loaded by default, so without this, `EVP_get_cipherbyname("rc4")` returns NULL and the Windows code-signing PVK path (which calls `EVP_rc4()` in `deploy_sign.cpp`) would also break silently. Added `OSSL_PROVIDER_load`/`OSSL_PROVIDER_unload` to `ssl.stubs`.
- Vendor the OpenSSL 3.6.1 headers (from Homebrew openssl@3) into `prebuilt/include/openssl/` replacing the 1.1.1v tree. Reorder `prebuilt/versions/openssl` so `3.4.1` is the active first line.
- Fix a template deduction failure in `deploy_sign.cpp` — OpenSSL 3 added `const` to every `i2d_*` function's first argument, which broke the single-parameter `i2d<T>()` helper. Decoupled into `i2d<F, T>()` so implicit non-const → const conversion happens at the call site.
- Gate `no-rc5 no-hw` in `build-openssl.sh` on 1.x-only (both were dropped from OpenSSL 3.x `Configure`).
- Separate commit: `compile-mac` was previously only signing `HyperXTalk.app` + bundles with Hardened Runtime. Loose executables (`server-community`, `lc-compile`, `lc-run`, `lc-compile-ffi-java`, `installer-stub`) landed unsigned, so running them from `_build/` on macOS 26 got SIGKILL'd via the libffi MAP_JIT path (AMFI ignores `allow-jit` without `CS_RUNTIME`). `package-mac-bin` already did the right thing for the `mac-bin/` copies — this just mirrors that block into `compile-mac`.

## Intentionally NOT in scope

- Rewriting the low-level `RSA_*` API (`RSA_size`, `EVP_PKEY_get1_RSA`, `RSA_private_encrypt`/`_public_encrypt`/`_private_decrypt`/`_public_decrypt`, `RSA_free`) in `mcssl.cpp` to `EVP_PKEY_CTX_*`. These are header-deprecated in 3.x but still exported from `libcrypto.3.dylib`, so they still link and work, just with warnings. Rewriting them changes the error model and key-type detection, so it deserves its own PR with its own test plan.
- iOS stubs (`ssl_ios.stubs`) — not touched because iOS isn't targeted on this fork anymore. If iOS support comes back, that file needs the same cleanup.
- Windows prebuilt tarballs — the `ssl.stubs` symbol rename applies to all platforms, but I haven't regenerated the bundled `openssl-1.1.1g-x86-win32-src` tree. Windows builds are a follow-up.
- FIPS provider support (nobody's asking).

## Verification (macOS arm64, macOS 26.4 SDK, Xcode 26, Homebrew openssl@3 = 3.6.1)

- `make compile-mac` completes with `** BUILD SUCCEEDED **` and signs cleanly.
- `nm -gU _build/mac/Debug/HyperXTalk.app/Contents/MacOS/revsecurity.dylib` confirms: old 1.x names (`_EVP_CIPHER_CTX_block_size` etc.) are **gone**; new `_get_*`/`_SSL_get1_peer_certificate` names are present; `_OSSL_PROVIDER_load`/`_unload` exported.
- Interactive smoke tests in `HyperXTalk.app` message box:
  - `encrypt \"hello\" using \"aes-128-cbc\" with key ... and IV ...` round-trips (default provider)
  - `encrypt \"hello\" using \"rc4\" with key ...` round-trips (legacy provider loaded)
  - `put URL \"https://www.google.com/\"` returns 78642 bytes, empty `the result` (exercises `opensslsocket.cpp` incl. the `SSL_get1_peer_certificate` rename)
- CLI smoke test via `server-community`: AES-128-CBC, RC4, and DES-EDE3-CBC all round-trip cleanly from a LCS script.

**What would help to finalize this PR**:
- @mwieder — you mentioned in #50 that you could verify the linux build. Would you mind checking that `build-openssl.sh` still downloads and builds OpenSSL 3.4.1 cleanly on linux, and that `make compile-linux` works? I've only touched the 1.x-vs-3.x gate on `no-rc5 no-hw` so it *should* be a no-op for you, but I haven't tested it myself.
- @emily-elizabeth — any code review you can spare, especially on the engine source changes in `engine/src/mcssl.cpp` and `engine/src/deploy_sign.cpp`.

Happy to push follow-up commits to address anything that comes up in review.

## Test plan (checklist)

- [x] macOS arm64: clean `compile-mac` build
- [x] macOS arm64: `revsecurity.dylib` symbol audit (old names gone, new names present)
- [x] macOS arm64: AES-128-CBC encrypt/decrypt round-trip
- [x] macOS arm64: RC4 encrypt (legacy provider path)
- [x] macOS arm64: DES-EDE3-CBC encrypt (legacy provider path)
- [x] macOS arm64: HTTPS fetch via `put URL`
- [x] macOS arm64: server-community runs a crypto smoke script
- [x] linux: `build-openssl.sh` fetches + builds 3.4.1
- [x] linux: `make compile-linux` succeeds
- [x] linux: runtime crypto smoke test


mwieder:
A few things here:
in thirdparty/libopenssl/libopenssl.gyp the line
'openssl_api_compat%': '0x30401000L',
needs to be the more generic
'openssl_api_compat%': '0x30400000L',

Do we need initialise_weak_link_ssl in revdb/src/ssl.cpp? I had errors compiling it and on thinking about it I'm not sure we want to be supporting outdated ssl formats.

Same for OSSL_PROVIDER_load in engine/src/mcssl.cpp if we're not on Windows. Linker was throwing an undefined error so I threw '#ifdef _WINDOWS' around that code block. Compiled and so far seems to run properly. I'm not getting a linker error on EVP_rc4 so it must be getting compiled from somewhere in the openssl structure.

Thanks for moving this forward. I had started on this a while back, ran into the RSA stuff and backed away. There's no easy migration path for that and it's a mess. Luckily the openssl code itself uses some of the deprecated code, so if and until they change that we should be ok.

Update: Ah, spoke too soon. Compiles properly but fails some key tests (both encrypt and url).

docmeth02:
thanks for the review.

`0x30401000L → 0x30400000L` nit is fair, will drop the floor in the next push.

the OSSL_PROVIDER_load call does need to run on mac and linux too, not just windows. it's not there for the windows code-signing path — it's there for user-land `encrypt using "rc4"` / `"des"` / `"blowfish"` etc. openssl 3 moved all those ciphers into the legacy provider which isn't loaded by default, so without the load call `EVP_get_cipherbyname("rc4")` returns null and every existing script that uses those ciphers gets `invalid cipher name` at runtime. verified on mac end-to-end: without it the rc4 round-trip fails, with it it works. your "compiles but encrypt tests fail" note after the `#ifdef _WINDOWS` wrap is basically the same breakage from the other side.

that said, whether we *should* still support those legacy ciphers going forward is a separate question and i don't have strong feelings — @emily-elizabeth that's probably your call. my vote for this PR is preserve backwards compat so existing scripts don't silently break, and deprecation can be a follow-up if you want it.

the undefined reference to `OSSL_PROVIDER_load` on linux is probably a stale `prebuilt/build/linux/` from an earlier 1.1.1 run. the branch moves 3.4.1 to the first line of `prebuilt/versions/openssl` so build-openssl.sh should fetch 3.x, but it happily reuses any existing build dir. can you `rm -rf prebuilt/build/linux` (or wherever your openssl output landed) and re-run? the symbol only exists in libcrypto 3.x, so if you're still linking a cached 1.1.1 libcrypto you'd see exactly that error. i'd bet the `revdb/src/ssl.cpp` compile issue is the same thing — that file is just the weak-link bootstrap revdb uses to reach libssl for TLS to mysql/postgres, nothing to do with legacy formats, so it should be fine once the underlying libs are actually 3.x.

which URL test is the one failing? happy to dig once i know which path.

mwieder:
Yeah, that's a good call. I *thought* I had cleared out the existing stuff, but maybe not. It's easy to do... invoking "make clean-linux" does the trick as a one-liner. Rebuilding now.

Also I changed lines 34-35 of prebuilt/scripts/build-openssl.sh to
if [ ! -d "${INSTALL_DIR}/../$OPENSSL_SRC" ] ; then
	echo "local openssl directory (${INSTALL_DIR}/../$OPENSSL_SRC) not found"
to eliminate some extra thrashing. Don't know if you want to add that to this PR or if I should submit it separately.

and the failed url test is the simple "put url "https//www.google.com". No point in doing any further digging yet.

emily-elizabeth:
Is this safe to merge in?

mwieder:
OK - I verified that there's no openssl 1.1.1 in my system. My base ubuntu system has openssl 3.0.1 but the build process only seem to link to local build files anyway. Here's what I'm seeing:

/usr/bin/ld: out/Debug/obj.target/engine/libkernel.a(mcssl.o): in function `load_crypto_symbols()':
/home/mwieder/Downloads/HyperXTalk/build-linux-x86_64/livecode/../../engine/src/mcssl.cpp:94:(.text+0x1e): undefined reference to `initialise_weak_link_ssl'
/usr/bin/ld: out/Debug/obj.target/engine/libkernel.a(mcssl.o): in function `InitSSLCrypt()':
/home/mwieder/Downloads/HyperXTalk/build-linux-x86_64/livecode/../../engine/src/mcssl.cpp:126:(.text+0x98): undefined reference to `OSSL_PROVIDER_load'
/usr/bin/ld: /home/mwieder/Downloads/HyperXTalk/build-linux-x86_64/livecode/../../engine/src/mcssl.cpp:127:(.text+0xac): undefined reference to `OSSL_PROVIDER_load'
/usr/bin/ld: out/Debug/obj.target/engine/libkernel.a(mcssl.o): in function `ShutdownSSL()':
/home/mwieder/Downloads/HyperXTalk/build-linux-x86_64/livecode/../../engine/src/mcssl.cpp:695:(.text+0x1760): undefined reference to `finalise_weak_link_ssl'
collect2: error: ld returned 1 exit status

I grepped for 'initialise_weak_link' and only got a hit on prebuilt/bin/Revolution.osx, so that's not encouraging. I'm getting libcustomcrypto built but not libcrypto. Initialise_weak_link_crypto doesn't throw the same error.

mwieder:
It seems that liblegacy should be linked but isn't. Haven't yet found the right incantation to make that happen.

docmeth02:
@mwieder applied your build-openssl.sh path fix in cefa4913, good catch.

the linker errors are a bit puzzling though. i just ran `weak_stub_maker.pl` against the current `ssl.stubs` locally and the generated .cpp has all four missing symbols — `initialise_weak_link_ssl` at line 2859, `OSSL_PROVIDER_load` at line 2360, `finalise_weak_link_ssl` at line 2483. so the stubs file and the generator are fine.

the fact that you grepped for `initialise_weak_link` across the entire tree and only found it in `prebuilt/bin/Revolution.osx` is the real clue — it means `libopenssl_stubs.a` either wasn't built or the generated stubs .cpp is stale/missing. `initialise_weak_link_crypto` and `initialise_weak_link_ssl` come from the same generated file, so if one is truly missing the other should be too. worth double-checking that you're not just seeing a different build error that scrolled past.

couple things to try:

1. after pulling this branch, make sure you ran `make config-linux` (not just `compile-linux`). `make clean-linux` wipes `build-linux-*` which has the gyp-generated Makefiles — without re-running config, there's no Makefile to drive the stubs generation action. `make all-linux` does both in sequence if you want a one-liner.

2. check if the stubs library actually has symbols: `nm build-linux-x86_64/livecode/out/Debug/obj.target/thirdparty/libopenssl/libopenssl_stubs/libopenssl_stubs.a 2>/dev/null | grep initialise_weak_link` — if that comes back empty, the gyp action that generates the .cpp from ssl.stubs didn't fire. look for the generated file at `build-linux-x86_64/livecode/out/Debug/obj.target/thirdparty/libopenssl/libopenssl_stubs/src/ssl.linux.stubs.cpp` and see if it exists / looks right.

3. if you're building from a directory that had previous builds, try a fully clean slate: `rm -rf build-linux-* _build/linux prebuilt/build prebuilt/lib && make all-linux`

@emily-elizabeth i'd hold off merging until mark's linux build is clean. the mac side is verified end-to-end but we should make sure linux isn't silently broken before this lands on main.

also dropped the api compat floor to 3.4.0 in the previous push (fdacb90c) per mark's earlier note.

docmeth02:
the liblegacy issue is that with `shared` in the Configure options, openssl 3.x builds the legacy provider as a separate `legacy.so` module instead of compiling it into `libcrypto.a`. revsecurity whole-archives the static lib so `OSSL_PROVIDER_load` is there, but the actual legacy provider implementation isn't — at runtime it tries to dlopen `legacy.so` from the build-time install prefix and silently fails.

pushed `6035b6d3` which adds `no-module` to the 3.x Configure options. that bakes all providers (default + legacy) directly into `libcrypto.a` as builtins, so they ship inside revsecurity and work at runtime without needing any external .so files. gated on 3.x only since the option doesn't exist in 1.1.1. you'll need a clean openssl rebuild after pulling — wipe whatever directory `build-openssl.sh` outputs to and re-run it so Configure picks up the new flag.

the `initialise_weak_link_ssl` linker errors are a separate issue though — that's `libopenssl_stubs.a` not being generated. did you get a chance to try the diagnostics from my earlier comment? (`make config-linux` + check `nm` on the stubs .a)

@emily-elizabeth this should unblock the linux side once mark confirms. holding off on "ready to merge" until then.

mwieder:
I ran weak_stub_maker.pl against ssl.stubs and none of the missing symbols are in the resulting .cpp file.
I did a make config-linux after make clean-linux, both after push 6035b6d3 and still the ssl_stubs.a file is not getting built. And thus getting the same linker errors as before.
I'm stumped.

mwieder:
Stupid, stupid, stupid. Never mind - I made a typo in some debugging code to see what was going on, and it failed silently and then not so silently. This (with the additions) is good to go now on linux. Apologies for the waste of bandwidth.

mwieder:
I'm finding that 'put url "https://www.google.com"' fails while  'put url "http://www.google.com"' succeeds. Is this just a linux thing? Can anyone verify that it's working on other platforms or did we break something along the way?

emily-elizabeth:
Seems to be working on macOS and Windows

image

image

mwieder:
Found it. Now just have to figure out why.

[3617888] Unable to load: SSL_get_peer_certificate
[3617888] Warning: could not load library: ./revsecurity

docmeth02:
@mwieder the `Unable to load: SSL_get_peer_certificate` is because that symbol is only a `#define` macro in OpenSSL 3.x (ssl.h:1810), not an actual exported function — the weak linker tries to dlsym it, fails, and bails. fix is up in #143, just removes the dead name from ssl.stubs and the version guard in opensslsocket.cpp. can you try the `fix-ssl-get-peer-certificate` branch and see if that clears it up?

mwieder:
Y'all are fast... stepped out of the room for a bit and came back to this! The writeup makes sense. Building now.

docmeth02:
@mwieder #143 is merged and i verified https is working locally after a clean rebuild. we also found two unrelated linux build issues that had crept into main (`cairo-features.h` missing FreeType defines, `libcurl.gyp` missing `-lnghttp2`) — those are fixed in #144. if you pull main and do a `make clean-linux && make all-linux` the `SSL_get_peer_certificate` error should be gone.

Issue #67, opened 2026-04-16
Clean up the macOS/Linux build so it stops depending on hacks


docmeth02:
Hey! After getting the macOS arm64 build green again (PR #52 + #53), a lot of the fixes ended up as workarounds rather than clean solutions. It would be nice to go through them one by one and make the build boring again. Opening this issue to discuss scope and priority before sending PRs.

Scope: **macOS and Linux**. Windows is out of scope from my side — I don't have a Windows machine to test on, so someone else would need to drive that.

### Todo list — things currently hacky

- [x] **libzip** — the vendored tree has 2005-era headers mixed with modern 1.10.1 source. The xcodeproj can't build it, so we hand-compile it in `build-mac-extras.sh` with a sed-patched `config.h`. Fix: re-vendor a clean libzip 1.11.x, regenerate the gyp, drop the manual build.
- [x] **ICU 58.2 living in `/tmp`** — `build-icupkg-mac-arm64.sh` builds ICU into `/tmp/icu58_arm64_build/` and then `build-mac-extras.sh` copies 5 `libicu*.a` out of it. If `/tmp` gets cleaned between steps the build silently breaks. Also 58.2 is from 2016. Fix: proper prebuilt script with a real install path, or bump to a current ICU.
- [x] **libpq / libmysql stubs** — empty archives so the linker is happy; real symbols come at runtime via `-undefined dynamic_lookup`. Works, but the moment someone actually uses dbpostgresql or dbmysql they'll hit a runtime error. Fix: ship real prebuilts, or teach the gyp to skip the link flag.
- [ ] **fetch-libraries.sh is bitrotted on macOS** — we bypass it entirely and call a bunch of `build-*-mac-arm64.sh` scripts manually. There's no single "build the deps" command anymore. Fix: a top-level `make prebuilt-mac` (and `prebuilt-linux`) that does the right thing.
- [ ] **libffi strips DWARF CFI directives by hand** — Apple's `as` rejects them, so `build-libffi-mac-arm64.sh` preprocesses `.S` files and sed-strips `.cfi_*` lines. Works, but surprising. Also no version pin on the vendored source.
- [ ] **libz compat hack** — `build-libz-mac-arm64.sh` compiles 7 files with `-include unistd.h` for older SDKs. Fine, but could just be a proper zlib 1.3.x vendor.

### Linux side
[...]

PR #70, opened 2026-04-16
icu: consolidate build into one script; install off /tmp

docmeth02:
## Summary

Merges the icupkg build and the five libicu*.a copies into a single script, and points it at a durable `prebuilt/build/icu-58-mac-arm64/` path instead of `/tmp/icu58_arm64_build/`. The old split between `build-icupkg-mac-arm64.sh` and `build-mac-extras.sh` silently broke whenever macOS cleaned `/tmp` between the two steps — issue #67's ICU checklist item.

Progress on #67.

## What's in here

One commit on `main`:

- Rename `prebuilt/scripts/build-icupkg-mac-arm64.sh` → `prebuilt/scripts/build-icu-mac-arm64.sh` and rewrite it. One pass installs `icupkg` into `prebuilt/bin/mac/` and the five `libicu{data,i18n,io,tu,uc}.a` into `prebuilt/lib/mac/`. Downloads are SHA256-verified (`2b0a4410…`) and extraction is idempotent across interrupted runs.
- Drop the ICU section from `prebuilt/scripts/build-mac-extras.sh` and its header bullet; renumber the stub-libs section.
- `BUILD1.md` section 5 retitled to reflect the new scope.
- `prebuilt/versions/icu`: first line corrected to `58.2` (matches what the script builds and what the checked-in `icudt58l.dat` was generated from). Attempted-bump history preserved as comments.

## Why 58.2 (not a bump)

`prebuilt/versions/icu` history makes it clear bumping is hazardous:
- 64.1 needs `distutils` (removed from Python 3.12).
- 65.1 coredumps during `Generating full lc-compile grammar`.
- 63.2 is the highest tried without a coredump.

So this PR takes the "real install path" option from the issue description and leaves the bump for a separate, careful PR.

## Verified on macOS arm64

- Fresh run installs to `prebuilt/build/icu-58-mac-arm64/`; `/tmp` never populated.
- Freshly built `icudt58l.dat` is **byte-identical** to the checked-in `prebuilt/share/icudt58l.dat` (SHA256 `549f2affdbd961742fbc07fd786808b938b3739b29cb42a661bbeaf3782e4b3f`) — proves the build is reproducible against the known-good data file.
- `prebuilt/bin/mac/icupkg --help` runs (ad-hoc signed, arm64).
- `libicudata.a` is 728 bytes — correct behaviour for `--with-data-packaging=archive`; the stub declares `_icudt58_dat` and the real data lives in `icudt58l.dat`.
- `make compile-mac` → `** BUILD SUCCEEDED **` with signed `HyperXTalk.app`.
- Script idempotent across two back-to-back runs.

## Out of scope

- **ICU version bump** — separate hazardous task (see above).
- **Linux / iOS / Android / Windows ICU** — those already work from checked-in prebuilts under `prebuilt/lib/{linux,ios,android}/`.
- **`.ao` / `.d` cruft in `prebuilt/include/unicode/`** (366 stray build artifacts) — pre-existing mess, its own cleanup.
- **Per-arch lib layout mismatch** between `prebuilt/libicu.gyp` (expects `lib/mac/<arch>/`) and the hand-maintained xcodeprojs (use flat `lib/mac/`) — pre-existing and independent of this change.

## Test plan

- [ ] macOS arm64 from a tree with no `prebuilt/build/icu-58-mac-arm64/`: `sh prebuilt/scripts/build-icu-mac-arm64.sh` runs clean.
- [ ] `ls prebuilt/bin/mac/icupkg prebuilt/lib/mac/libicu{data,i18n,io,tu,uc}.a` — all 6 artifacts present.
- [ ] `ls /tmp/icu58_arm64_build 2>&1` returns "no such file".
- [ ] `make config-mac && make compile-mac` → build succeeds, `HyperXTalk.app` runs.
- [ ] Smoke test: launch the built app; `put the internationalNumericLocale` in the Message Box returns a locale-formatted number.


emily-elizabeth:
@copilot resolve the merge conflicts in this pull request

PR #98, opened 2026-04-19
allow specialFolderPath(documents) on linux

mwieder:
This patch adds support for the `documents` keyword to the `specialFolderPath` function using the same (simple) implementation approach as the `desktop` keyword, i.e. the users home folder is appended with `/Documents`. Note that this, like the other specialFolderPath() options, will work with English names for the folders. Not clear if there's an easy way to localize these.


mwieder:
Note that this is a fix for bugzilla #15094.

Issue #82, opened 2026-04-16
Update libiodbc from 3.52.5 to 3.52.15


emily-elizabeth:
## libiodbc

**Current vendored version:** 3.52.5
**Target version (per `prebuilt/versions/libiodbc`):** 3.52.15
**Location:** `thirdparty/libiodbc/`

The `prebuilt/versions/libiodbc` file already targets 3.52.15. This is a minor version bump within the 3.52.x stable series.

### Update steps
1. Download iODBC 3.52.15 from http://www.iodbc.org
2. Replace `thirdparty/libiodbc/` source
3. Update `thirdparty/libiodbc/ORIGIN`
4. Rebuild and verify ODBC database connectivity

emily-elizabeth:
macOS is 3.52.16

mwieder:
Why was this reopened? 3.52.16 is the current build version.

emily-elizabeth:
I have them open to check on all the platforms to make sure they're all on the same version

mwieder:
OK. Linux is on 3.52.16

emily-elizabeth:
Windows is on 3.52.16 as well, so time to close this

Issue #81, opened 2026-04-16
Reconcile libfreetype partial update (2.10.4 / 2.14.3)


emily-elizabeth:
## libfreetype

**Situation:** The source tree contains two sets of headers:
- `thirdparty/libfreetype/include/freetype/freetype.h` — **2.10.4** (old)
- `thirdparty/libfreetype/include/include/freetype/freetype.h` — **2.14.3** (new)

This suggests a partial update was started but not completed. The `prebuilt/versions/libfreetype` file targets 2.14.3.

### Update steps
1. Determine which header path the engine and build system actually consume
2. Remove the stale 2.10.4 headers
3. Ensure the full 2.14.3 source is in place throughout `thirdparty/libfreetype/`
4. Update `thirdparty/libfreetype/ORIGIN` to reflect 2.14.3
5. Rebuild and verify font rendering

Issue #80, opened 2026-04-16
Update libpng from 1.6.26 to 1.6.56


emily-elizabeth:
## libpng

**Current vendored version:** 1.6.26
**Target version (per `prebuilt/versions/libpng`):** 1.6.56
**Location:** `thirdparty/libpng/`

The `prebuilt/versions/libpng` file already targets 1.6.56 (current upstream). The source tree just needs to be updated to match.

### Update steps
1. Download libpng 1.6.56 from http://www.libpng.org/pub/png/libpng.html
2. Replace `thirdparty/libpng/` source
3. Update `thirdparty/libpng/ORIGIN`
4. Rebuild and verify PNG image handling

emily-elizabeth:
updated macOS to 1.6.46

emily-elizabeth:
Windows updated to 1.6.46

PR #99, opened 2026-04-19
Restore Linux build on Ubuntu 24.04

docmeth02:
## Summary

The Linux build has been broken for a while due to Python 3 incompatibilities,
deleted source files, deprecated dependencies, and outdated bundled libraries.
This PR gets it compiling again on Ubuntu 24.04.

- Fix Python 3 compatibility in build scripts (bytes/str, map → list)
- Remove deprecated GNOME VFS dependency, use xdg-open instead
- Fix ICU and zlib prebuilt build scripts for modern glibc
- Fix include guard ordering in database driver headers
- Restore accidentally deleted CEF source files (libbrowser + revbrowser)
- Link against system libmysqlclient / libpq on Linux instead of outdated bundled versions

Requires `libmysqlclient-dev` and `libpq-dev` installed on the build machine.

Tested: `make config-linux-x86_64 && make compile-linux-x86_64` completes
successfully on Ubuntu 24.04 with GCC 13.

## Known follow-ups (will be addressed in this PR before merging)

- build-libz.sh output path is Linux-only (doesn't affect other platforms
  unless prebuilts are rebuilt from source)
- Dead `MCuselibgnome` global can be removed
- 4 remaining `map()` calls in make.py need `list()` wrapping
- Relates to #81 (freetype), #82 (libiodbc), #80 (libpng)

Screenshot from 2026-04-19 11-48-42


docmeth02:
@mwieder since you are working on linux, can you check this doesn't break any of your work etc?

mwieder:
It's close. A couple of things to get it going again, all in the buildicu.sh script. Firstly, the code to fetch and validate the sha code is commented out because the shasum file for icu version 58 doesn't have proper checksums for the icu file we're downloading (and I had to bump icu to 58.3 because 58.2 is no longer available). We could possibly download one of the Fedora builds and thus verify the sha code, but I commented out your resurrection of the code with "if [ ${ICU_VERSION_MAJOR} != 58 ] ; then" and moved on. Once we can update the icu library this won't matter. But I'm dismayed by the rather tight coupling of libfoundation and the icu api. The unicode project has a habit of relabeling things and making breaking changes often.
Secondly, your clever code for monkeypatching the <xlocale.h> code fell afoul of whitespace differences. I got it working thusly, and it appears safe to change any occurence of xlocale.h.

	# Fix for modern glibc (2.26+) where xlocale.h was removed
	find ${ICU_SRC} -name *.cpp -o -name *.c | xargs grep -l "xlocale.h" | while read f; do
		sed -i.bak 's/xlocale.h/locale.h/g' $f && rm -f "$f.bak"
	done

After that it's compiling and running properly. I'm getting several (4 sets of 4 instances) warnings about "SyntaxWarning: invalid escape sequence '\.'" but for the moment they're just warnings - due to be errors in some future python version, so we should eventually fix them, but there's no indication of where they're coming from other than line numbers. And maybe this is just my build if nobody else sees them.

mwieder (approved):
Thanks for this - looks good. I would have submitted the patch but I'm not clear on how to do that with a pull request in progress. Just comment? or issue a PR to your repo?

mwieder (approved):
This is great - those are indeed the line numbers I was seeing. How did you manage to find these? My grep-fu didn't catch them.
grep -r "\\\." * 
didn't do the job.

docmeth02:
thanks @mwieder for checking and all your help! 

the gypi files were evaluated as string and those regex were not properly escaped, i had to search for a while too ;)

Yup when i push to the source branch in my fork it automatically amends the new commits here, so creating a PR at the forked repo would be the way to go getting your changes into the PR draft 👍

Issue #123, opened 2026-04-22
Replace CEF with WebKitGTK on Linux


docmeth02:
CEF on Linux is showing its age — we're shipping Chromium 74 from 2019, the prebuilts are 217MB, and it only works on x86/x86_64. Windows already moved to WebView2, macOS uses native WKWebView. Linux should follow the same pattern with WebKitGTK.

The `libbrowser` factory/strategy architecture already supports swapping backends — we just need a new `MCWebKitGTKBrowser` implementing the `MCBrowser` interface.

What this gets us:
- System WebKitGTK instead of bundled Chromium (no more 217MB prebuilt blob)
- Always up to date with distro security patches
- Browser widget support on ARM Linux for the first time
- No more separate cefprocess executables

Blocked by #122 (GTK 2 → 3 migration) — WebKitGTK requires GTK 3.

mwieder:
Oooooh... I've been dreaming of this for ages.

PR #125, opened 2026-04-23
GTK 2 to GTK 3 migration

### LINKED ISSUE #122: Migrate engine from GTK 2 to GTK 3
The engine currently uses GTK 2 (`libgtk-x11-2.0.so.0`) loaded via dlopen through `engine/src/linux.stubs`. GTK 2 has been EOL since 2020 and blocks us from using modern Linux libraries that require GTK 3 — most notably WebKitGTK for replacing CEF in the browser widget (#83).

The GTK 2 usage is mostly contained to:
- `engine/src/linux.stubs` — dlopen symbol table
- `engine/src/native-layer-x11.cpp` — GdkRegion, widget embedding
- `engine/src/lnxgtktheme.cpp` / `lnxgtkthemedrawing.cpp` — theme rendering
- `engine/src/lnxans.cpp` — file/print dialogs
- `engine/src/lnxdcs.cpp` — display/clipboard

Main migration points:
- `GdkRegion` → `cairo_region_t`
- GTK 2 paint/draw APIs → GTK 3 CSS-based rendering
- Deprecated widget APIs (`gtk_object_sink`, `gdk_drawable_*`, etc.)
- Stub table: `libgtk-x11-2.0.so.0` → `libgtk-3.so.0`

This unblocks #123 (WebKitGTK browser widget replacement).

docmeth02:
Work in progress migration from GTK 2 to GTK 3. Closes #122.

The engine was using bundled GTK 2.10 headers from 2006 and dlopening libgtk-x11-2.0.so.0. This switches to system GTK 3 headers via pkg-config and updates all the API calls that changed or were removed between the two versions.

Most of the codebase is ported — type declarations, region handling, native layer embedding, display context, colormap removal, drawing operations moved to cairo. The C++ standard was bumped to C++14 for GLib compatibility.

Still working through lnxgtkthemedrawing.cpp and lnxgtktheme.cpp — the Mozilla-derived theme rendering code that uses gtk_paint_* extensively. That's the bulk of what's left.

Unblocks #123 (WebKitGTK browser widget).


docmeth02:
update: hit a roadblock with the engine crashing on wayland systems but got it sorted — GDK 3 defaults to the wayland backend which doesn't play nice with the x11-specific code throughout the engine. forced x11 backend early in startup and cleaned up the last GTK 2 stubs that were blocking library loading.

this now compiles and starts on a wayland system (via xwayland). all GTK 3 libraries load, the x11 display opens, windows are created and the IDE reaches its main loop with extensions loaded.

will need extensive testing from here though — pretty much all user interface code runs through the upgraded library. every window, dialog, drawing operation, theme rendering, clipboard interaction and input handling goes through GTK/GDK 3 now. the theme drawing in particular is a complete rewrite from gtk_paint_* to gtk_render_* so i'd expect visual glitches and possibly crashes in less common code paths.

mwieder:
Thanks for moving this forward. Gtk2->Gtk3 has some nasty compatibility breaking api changes. I've stuck my toe in the water and backed out pretty much immediately.

docmeth02:
update: ran the full LCS test suite against the gtk3 build — 47,200 tests, 11 failures, all pre-existing and unrelated to the migration. also wrote a stub symbol audit script that checks every required dlsym in linux.stubs resolves against the system gtk3/gdk3/glib/pango libraries — 492 out of 495 pass, the 3 missing are from libesd which is dead.

found and fixed a few theme init bugs the testing surfaced. moz_gtk_init() was never being called so the proto widget container didn't exist when the ensure_*_widget() functions tried to add to it. also gtk_widget_style_get was getting NULL output pointers for checkbox/radio metrics which corrupts the heap under gtk3. and GtkMenuItem was being parented by GtkFixed instead of a proper GtkMenu. windows render correctly now — no more black screens.

remaining issue is that CEF (chromium 74) corrupts the heap on any gui interaction because it was built against gtk2 and its GtkSocket embedding breaks under gtk3. this blocks interactive testing on the gtk3-migration branch alone — needs #142 (webkitgtk replacement) merged on top to fully test. added a widget test stack in _test_stacks that exercises all themed control types once that's possible.

PaulMcClernan:
Tom Perry at openxtalk.org also updated his Linux build to link to GTK3, but he **static** embedded GTK3, which seems like that could eventually be the same problem popping up down the road. His source is on his MEGA file share account (link to download source in OXT forums / OXTLite).
Thanks for this. IIRC GTK3+ Is needed in order to switch the browser widget from CEF to GTK WebKit. Hopefully it will also makes using Extension / LCB FFI with other GTK3+  stuff work without crashing engine.

PaulMcClernan:
Yes libesd is long dead (along with 'Enlightenment'). Tom replaced that in his Linux build too (not sure with what, mPlayer?), as well as the crashing when switching to/from certain System-wide UI themes (may have been old-GTK related).
Here's his Linux source: https://mega.nz/folder/bDQnFD4K#wCf6k9_wRx-l5IY0saZQzg/folder/WeQikLCT

mwieder:
I think it's fine to ignore CEF errors on linux. The browser stuff hasn't worked there in dog years. And building the cef libraries is the thing that takes the most time in a build. Maybe we can just make a linux build that doesn't have the lbbrowser stuff or any other CEF-related work.
IIRC Tom was also going to embed glib in his work. Don't know if that happened, and it doesn't seem like a better idea than embedding the GTK libraries either.

emily-elizabeth:
This work has been completed

mwieder:
I had thought that all the POLL and WAIT messages being spewed to stderr were just for debugging, but I see they're still there in the merged main branch now.

emily-elizabeth:
Yeah, that was my bad. It'll be removed before we ship

PR #143, opened 2026-04-24
fix SSL_get_peer_certificate breaking revsecurity on linux

docmeth02:
SSL_get_peer_certificate is a macro in OpenSSL 3.x, not an exported symbol — the weak linker fails to dlsym it and bails, killing the entire ssl module. removes the dead 1.x name from ssl.stubs and the version guard in opensslsocket.cpp, consistent with how the EVP renames were handled in #53.

fixes the `Unable to load: SSL_get_peer_certificate` / `could not load library: ./revsecurity` error @mwieder reported in https://github.com/emily-elizabeth/HyperXTalk/pull/53#issuecomment-4315148983.


mwieder:
Verifying the fix.

PR #144, opened 2026-04-24
fix linux build: cairo FreeType and nghttp2 linking

docmeth02:
two issues preventing a clean `make compile-linux` on current main:

- `cairo-features.h` was generated on macOS and committed with Quartz backends but without `CAIRO_HAS_FT_FONT` / `CAIRO_HAS_FC_FONT`. `revpdfprinter_lnx.cpp` needs `cairo_ft_font_face_create_for_ft_face` which is gated behind that define, so compilation fails. fix: gate platform backends with `#if defined(__APPLE__)` / `#if defined(__linux__)` conditionals.

- prebuilt `libcurl.a` was built with HTTP/2 support (`nghttp2`) but `libcurl.gyp` didn't link `-lnghttp2` on linux, so `server-community` fails to link with undefined `nghttp2_*` references. fix: add `-lnghttp2` to the linux link_settings.

verified: clean `make config-linux && make compile-linux -j4` succeeds, `server-community` links and runs, HTTPS fetch via `put url "https://www.google.com"` returns data.


mwieder:
Not quite sure what you're trying to do here. libcurl on linux is generated without nghttp2. Trying to link cairo with it just results in a 'not found' linker error. Wouldn't be a bad idea to add it, but that requires a dependency we'd have to add and build separately.

cairo-features.h is regenerated by meson, so the change you're suggesting gets overwritten at each build cycle. I wasn't getting build errors, but your writeup looks ok, so to keep the FT/FC font changes and prevent the overwriting I made the following patches:

in thirdparty/libcairo/src/cairo-features.h:

#if defined(__APPLE__)
#define CAIRO_HAS_QUARTZ_FONT 1
#define CAIRO_HAS_QUARTZ_IMAGE_SURFACE 1
#define CAIRO_HAS_QUARTZ_SURFACE 1
#endif

#if defined(__linux__) || defined(__ANDROID__)
#define CAIRO_HAS_FT_FONT 1
#define CAIRO_HAS_FC_FONT 1
#endif

in prebuilt/scripts/build-libcairo.sh
if [ PLATFORM == mac ] ; then
	CAIRO_HAS_QUARTZ_FONT=1
	CAIRO_HAS_QUARTZ_IMAGE_SURFACE=1
	CAIRO_HAS_QUARTZ_SURFACE=1
elif [ PLATFORM == linux ] ; then
	CAIRO_HAS_FT_FONT=1
	CAIRO_HAS_FC_FONT=1
elif [ PLATFORM == android ] ; then
	CAIRO_HAS_FT_FONT=1
	CAIRO_HAS_FC_FONT=1
fi

and
#copy cairo-features.h
#cp -u ${BUILDDIR}/${ARCHIVE_DESTINATION}/build/src/*.h ${INCLUDE_DIRECTORY}

Compiles and seems to run without problems.
Am I missing something here?

docmeth02:
thanks @mwieder, i reverted the PR for now, i'll do a full freshly checked out build to confirm this was just a fluke caused by the GTK and WEBKITGTK changes creeping into my worktree 👍

mwieder:
It's more than just reverting the PR, I did indeed get cairo errors if I didn't make the changes above in cairo-features.h. Maybe a cross-platform thing?

PR #147, opened 2026-04-24
Implement mouse event handlers in defaultscript

PaulMcClernan:
Added mouse event handlers for mouseDoubleUp, mouseDown, mouseLeave, and mouseEnter to default scripts just to make using the SVGPath widget like a button a little bit easier.


mwieder:
@PaulMcClernan good to see you here.

PR #164, opened 2026-04-29
Features/toolbars

emily-elizabeth:
macOS and Windows toolbars


macMikey:
my head is spinning from all the stuff that's going in

emily-elizabeth:
Gotta keep the people happy 😄  Most of it is stuff I've always wondered why it wasn't in the IDE to begin with

PR #175, opened 2026-05-01
Feature sql date

mwieder:
Implements the sql date at the same level as the internet date, etc.
Allows conversion between formats.
Also solves the existing LC bug #24798


emily-elizabeth:
How is it used? I'll write a documentation entry for it

mwieder:
Thanks - I forgot to do the docs. Same as any other date - you can use internet date as a template.
put the sql date -- returns YYYY-MM-DD HH:MM:SS
put the seconds into xdate; convert xdate to sql date -- returns the sql format
put the sql date into x; convert to dateitems -- converts the other way around
..and similar to the internet date, there's no long/short version - it all comes out the same

Along the way I noticed that LC has a bug displaying the long internet date, so I fixed that as well.

macMikey:
mark is my other hero

emily-elizabeth:
docs created

PR #195, opened 2026-05-08
Feature/battery status

emily-elizabeth:
Get the battery level and power source

`put powerSource() into tSource `
`put batteryLevel() into tLevel `


engine/src/w32-battery.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    if (t_status.BatteryLifePercent == 255)
+    {
+        // No battery present or status unavailable.
+        r_level = -1;
+    }
+    else
+    {
+        r_level = (integer_t)t_status.BatteryLifePercent;
+        if (r_level < 0)   r_level = 0;
```
mwieder, line 46: Can the battery level actually be <0 or >100? Thanks, Microsoft.

build-mac/livecode/engine/engine-common.xcodeproj/project.pbxproj
```diff
@@ -386,7 +386,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "exec perl ../util/encode_version.pl -275c7785d4a74bc3950152e30e0879454d1e3da7 . \"${SHARED_INTERMEDIATE_DIR}\"\nexit 1\n";
+			shellScript = "exec perl ../util/encode_version.pl -3bf17f3115293baf1cb1416bd8c47110ef3ea164 . \"${SHARED_INTERMEDIATE_DIR}\"\nexit 1\n";
```
mwieder, line 389: Someday I'd like to get rid of our reliance on perl, which I tend to regard as a write-only language.

mwieder (approved):
I'm not going to comment on the osx sections, I don't think I'm particularly qualified to do that at this point. Those sections look logical to me but that's as far as I'm going to go. The linux and Win sections look good.

PR #198, opened 2026-05-08
Feature/global hotkeys

emily-elizabeth:
Set a hotkey to be used when your application is in the background.

`registerHotkey keyString, handlerName`
`unregisterHotkey keyString`
`unregisterAllHotkeys`

There is a test stack `_test_stacks/GlobalHotKeys.hyperxtalk`


emily-elizabeth:
[GlobalHotKeys.hyperxtalk.zip](https://github.com/user-attachments/files/27544580/GlobalHotKeys.hyperxtalk.zip)

engine/src/lnx-hotkey.cpp
```diff
@@ -0,0 +1,402 @@
...
+//   MCHotkeyDispatchFired() is always called from the correct thread.
+//
+// XGrabKey registers with Num Lock (Mod2) and Caps Lock (LockMask) masked
+// out so the hotkey fires regardless of those lock states.
+//
+// Requires: -lX11  (already linked in the Linux desktop build)
+//           GLib / GDK (already linked)
+//
+
+#include "prefix.h"
```
mwieder, line 30: X11 and Gdk have conflicting typedefs.
prefix.h includes sysdefs.h, which already has typedefs that conflict with X11 typedefs for
Atom
Window
Drawable
PixMap

Switching the load order doesn't help

mwieder (changes requested):

mwieder:
String conflicts in parseerrors.h and executionerrors.h.

mwieder:
kernel.gyp needs to link two libraries.
This starts at line 321
							'libraries':
							[
								'-ldl',
								'-lpthread',
								'-lcups',
								'-ldbus-1',
								'-lX11', # added
								'-lgio-2.0', # added
							],

emily-elizabeth:
Any further testing of this?

mwieder:
After spending most of the day on this... still doesn't compile.
glib-unix.h isn't in my path, but I have it in /usr/include/glib-2.0/. I don't know if that's a standard across distros, but just in case it is, I added that to kernel.gyp 'include-dirs'. That got me to undefined macros, so I added #include <glib/gmacros.h> to the lnx-hotkey.cpp headers. That didn't relieve the undefined GLIB_AVAILABLE_IN_2_30, so I also added <glib/glib-visibility.h>. That left me with undefined G_GNUC_BEGIN_IGNORE_DEPRECATIONS. Added GLIB_DISABLE_DEPRECATION_WARNINGS to kernel.gyp's defines but that didn't help.

And mind you that's just to get the code to compile. I'm still not sure what this feature does or why it's useful.

emily-elizabeth:
> And mind you that's just to get the code to compile. I'm still not sure what this feature does or why it's useful.

You ever have a program installed where it sits in the background and you press CTRL+ALT+B to bring it to the front? It allows your applications to register keyboard shortcuts to run commands without the program being the front most application.

emily-elizabeth:
> After spending most of the day on this... still doesn't compile. glib-unix.h isn't in my path, but I have it in /usr/include/glib-2.0/. I don't know if that's a standard across distros, but just in case it is, I added that to kernel.gyp 'include-dirs'. That got me to undefined macros, so I added #include <glib/gmacros.h> to the lnx-hotkey.cpp headers. That didn't relieve the undefined GLIB_AVAILABLE_IN_2_30, so I also added <glib/glib-visibility.h>. That left me with undefined G_GNUC_BEGIN_IGNORE_DEPRECATIONS. Added GLIB_DISABLE_DEPRECATION_WARNINGS to kernel.gyp's defines but that didn't help.

I'll ask Claude about this

mwieder:
So this would allow you to assign a hotkey combination that works for any app? or is it stack-specific?
I can't say that this is a feature I've longed for.

emily-elizabeth:
> So this would allow you to assign a hotkey combination that works for any app? or is it stack-specific? I can't say that this is a feature I've longed for.

It would specific to your application. It's not a well-known feature, but it could be useful to some. I'm just about done working on something and then I'll ask Claude, give me like 10-15 minutes

mwieder:
Don't rush. Got another iron in the fire.

emily-elizabeth:
> The cascade of missing macros is the classic symptom of missing glibconfig.h — it's not in /usr/include/glib-2.0/, it lives under the lib path (something like /usr/lib/x86_64-linux-gnu/glib-2.0/include/). Without it, all the visibility and deprecation macros fall apart. The right fix is to pull both paths from pkg-config rather than hardcoding just the one.
> 
> Let me look at the branch before suggesting anything:
> 
> There it is. The Linux include_dirs uses pkg-config for dbus but not for glib — so it only gets one path, missing the generated glibconfig.h that lives under the lib tree. The fix is the same pattern already used for dbus:
> 
> Also revert the manual workarounds the tester added (the hardcoded path and the extra includes), since pkg-config supersedes them:
> 
> Good — the tester's workarounds didn't make it into the committed files, so nothing to revert. Commit the gyp fix:
> 
> That's pushed. The tester will also need to do a full gyp regenerate + rebuild for this to take effect. Once the include cascade is resolved, you'll find out what the next hurdle is — but at least it'll be an actual logic error rather than header plumbing.

mwieder:
Getting deep in the weeds on this. Part of the problem, I think, is that we are prepackaging some headers in thirdparty/headers/linux. And that includes stale headers for gdk and glib. They take precedence over the builtin system headers that the config option points to, partially because we've said to use them in engine.gypi. If I comment that out I get an error because the gdk headers aren't found. I can worm my way through that to the next error, but this is getting frustrating. There are some serious system-level changes that need to be made to get this acceptable.

emily-elizabeth:
> we are prepackaging some headers in thirdparty/headers/linux. And that includes stale headers for gdk and glib. 

Is there something we can do so that they use the system ones?

> There are some serious system-level changes that need to be made to get this acceptable.

I'm all ears 😸

mwieder:
To start with, this is going to interact with the work docmeth02 is doing to update the gtk/gdk libraries. I don't know if there's a similar hack for the gdk libs like the config hack for the glib libs, but if there is, it's no doubt going to run up afoul of the new gdk/gtk stuff. The thirdparty/headers/linux stuff is outdated but works. It's going to be major work to reconfigure the build process *not* to use the contents.
I'd say pass on this for now and reserve it for pulling into the repo in the future.

emily-elizabeth:
Sounds good.

bwmilby:
If it works for Mac and/or Win, could the work just be split?  Merge everything for the working OS into one commit and document as available for that OS.  Pull all of the in-process work for Linux into a fresh branch based off of what gets into main for later work once the other display manager stuff is done on Linux.

mwieder:
Possibly. Wouldn't be my favorite way to move forward due to piling up technical debt, and I also see lots of intermediate files that shouldn't be in this PR.

bwmilby:
We probably should be doing squash commits before merging PRs.  Doing it this way, we could remove things that shouldn't there and then the squash will get rid of the initial add.  Same concept for removals that need to be reversed.

emily-elizabeth:
All the conflicts have been removed from the PR. If the feature isn't ready for all three platforms, then it can wait until we can release it for all three.

emily-elizabeth:
Too many conflicts, I'm going to close this and there is a new PR for this feature

PR #199, opened 2026-05-08
Feature/bring application to front

emily-elizabeth:
This will cause the application to be brought to the front. Works handy with the global hot keys


emily-elizabeth:
`bringApplicationToFront`

mwieder (approved):
Looks good.

engine/src/lnx-activate.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "objdefs.h"
+#include "parsedef.h"
+
+#include "stack.h"
+#include "globals.h"
+#include "lnxdc.h"
+
+void MCLinuxActivateApplication(void)
+{
+    if (MCdefaultstackptr == nullptr)
```
mwieder, line 34: This should read
    if (MCdefaultstackptr == (MCStackHandle)nullptr)
        return;

in order to compile. The compiler is very picky about this.

engine/src/lnx-activate.cpp
```diff
@@ -0,0 +1,45 @@
+/* Copyright (C) 2003-2015 LiveCode Ltd.
```
mwieder, line 1: Sigh... I missed a couple of things.
The header for lnx-activation.cpp is still Livecode rather than HyperXTalk.

emily-elizabeth: I'll have Claude take care of that across the whole repo

engine/src/lnx-activate.cpp
emily-elizabeth: fixed

PR #200, opened 2026-05-09
Feature/field input validation

emily-elizabeth:
Allows for you to validate the information being entered into the field, based on the input type of the field


emily-elizabeth:
[FieldInputValidation.hyperxtalk.zip](https://github.com/user-attachments/files/27544017/FieldInputValidation.hyperxtalk.zip)

mwieder (commented):
At the moment it's hard to tell where things are going wrong, but I'll put more time into it. Evaluation seems overly lax. Symptoms:
evaluating email is ok with just "me"
evaluating pattern of date is ok with "hello world"

emily-elizabeth:
Did you click the buttons first?

mwieder:
OK - I thought I had, but it appears I didn't.
Now I did and I find the date field still accepts 99/99/9999 and the email field happily accepts qwerty::///@www....com.\\.www
But more of a problem is that once the validation fails and the red X shows up you have to correct the error. Otherwise the stack is hung waiting for acceptable text: you can't edit anything else.

emily-elizabeth:
Yeah, that's a feature (not being able to leave the field until it passes validation, you can change it in the script though) I wish I had a Linux machine to test on

mwieder:
OK. I must have missed that in the docs then. Would that be handled in a fieldValidationFailed handler?

emily-elizabeth:
yeah, in the card script

engine/src/exec-interface-field.cpp
```diff
@@ -1017,6 +1019,236 @@ void MCField::SetHintText(MCExecContext& ctxt, MCStringRef p_string)
...
+            return t_err;
+        }
+        return nullptr;
+    }
+
+    if (MCStringIsEqualToCString(m_input_type, "url", kMCStringOptionCompareCaseless))
+    {
+        // Host must contain a dot with at least 2 characters after it (TLD),
+        // and at least 1 character before it. Path/query/fragment are optional.
+        if (!s_regex_match(t_utext, "^(https?|ftp)://[^\\s/?#]+\\.[^\\s/?#]{2,}([/?#]\\S*)?$"))
```
mwieder: This is limited to just http(s) and ftp. Url's can be many more protocol types.

emily-elizabeth:
> the date field still accepts 99/99/9999 

I think that is due to the regex that is behind the date field, you can change it in the button script

emily-elizabeth:
> the email field happily accepts qwerty::///@www....com.\.www

I updated the regex and now that no longer validates

mwieder:
To resolve the build conflicts:
Do the Windows dlls need to be in the git repo? Aren't they rebuilt when necessary?
Same for the Standalone binary, which should be generated by a new build.
And the sqlite file. Possibly the same for the other two, maybe regenerated by a config call?
I suggest adding them to the .gitignore file to solve the issue.

emily-elizabeth:
Yes they get recreated every build, but I have to have them in the repo so they get to the Mac so we can build for Windows on macOS

emily-elizabeth:
api.sqlite is already in the gitignore, but it seems to be ignoring it and being transferred anyways

mwieder:
Yeah, I see that. That's weird

mwieder:
So the problem now with the dlls is that there is already a newer build in the main branch. In order to merge this pull request you have to resolve that, i.e., accept the existing ones and ignore the older versions in this PR. And that will continue to keep happening with future PRs if they're in the repo.

emily-elizabeth:
Yup, but I don't know another way to get the files to the Mac for the build. Now that they're there I guess we could ignore them again

mwieder:
Can the build process for the mac be modified to create those files as necessary?

emily-elizabeth:
Well they have to be compiled, so not sure that Mac can compile Windows code

emily-elizabeth:
The conflicts aside, can I merge this?

bwmilby:
Adding things to the ignore file does not impact things already tracked.  Once tracked, you need to either remove it from the cache (which leaves it on your machine) or be sure to not include it when you commit unrelated things.  This is where GitHub desktop is nice - easy to see files that shouldn't be committed (api.sqlite) and just tell it to revert to clean your branch.

From memory, I think it is better to rebase than to fix conflicts by merge.  Probably best to squash first and then rebase.  This will make the history much easier to understand too.  I've already fixed PRs in flight by merge, but will change going forward.

bwmilby:
I created a squashed commit of this PR on my fork that you can view.
https://github.com/bwmilby/HyperXTalk/tree/bwm-field-input-validation
I compiled/tested my branch.  I removed all of the api.lcdoc files from being tracked in my branch.
If you like, I can open a PR on it and this one could be closed.

emily-elizabeth:
created a clean version of this PR
https://github.com/emily-elizabeth/HyperXTalk/pull/345

PR #201, opened 2026-05-09
Feature/ide autocomplete

emily-elizabeth:
Autocomplete in the Script Editor

image


emily-elizabeth:
You may have to add Linux to the platform guard on line 404 of "ideautocompletelibrary" if the popup steals the focus on you

mwieder (approved):
This mostly works. On quitting HXT I get a segfault, but it seems harmless.
Also running glx2 instead of the builtin script editor I get "Tried again" for each keypress. Again seems harmless.

emily-elizabeth:
Does that change make the segfault go away?

mwieder:
Still get the segfault but it's harmless. We're good.

PR #232, opened 2026-05-19
Feature/row tags

emily-elizabeth:
Assign data to a row in a listbox. This data travels with the row and is saved with the stack


emily-elizabeth:
Feature is already in HyperXTalk

PR #233, opened 2026-05-19
Can't auto merge.

PaulMcClernan:
These are the changes needed to 'debrand' the dictionary so that "LiveCode Script" becomes "Script" and "LiveCode Builder" becomes "Extension Builder". The updated API.sqlite file included has API categories to match. The revIDElibrary had to be updated to use the rebranded names as well. I tried to keep the changes to a minimum, but might have to do a manual diff compare.


PaulMcClernan:
Will manually merge so there's no conflicts.

PR #235, opened 2026-05-19
Dictionary debranding and add 'Guides' back

PaulMcClernan:
I think I got these so that it's just the minumal differences to change dictionary 'branding' without affecting anything else Emily is doing with the rest of the libraries (revIDELibrary is particularly sensitive to changes).
The Dictionary UI binary stack also needs to be changed to put the 'Guides' tab button back, because currently there is no way for users to view the guides.


PaulMcClernan:
I'm going to go ahead and merge this, can always do a reversion later if need be.

PR #237, opened 2026-05-19
OK I think I got these so that it's just the minumal differences to change dictionary 'branding' without affecting anything Emily is doing with the rest of the libraries (revIDELibrary is particularly sensitive to changes). The Dictionary UI binary stack also needs to be changed to put the 'Guides' tab button bac, because currently there is no way for users to view the guides.

PaulMcClernan:
Replace revDictionary stack to bring back 'Guides' tab Also merged in small external dictionary behavior script for convenience (it was a small script file and just made editing it more of a pain).
I had a **CARD**-level search field in there as a work-around for Linux Browser Widget keyboard-focus problems, but took it out since it may not be needed with the revised Browser Widget (I suspect that problem is CEF related).

You can see in this screenshot the 'Debranding' in the in-line drop down API menu.
Screenshot 2026-05-19 at 9 02 00 AM
Screenshot 2026-05-19 at 9 10 18 AM


BerndN:
Paul,

I think that field "SearchKeyWords" at the top of the dictionary is redundant since there is another search field for the dicts. And it is awkward to use. (you have to tab into it, one can not click into it to edit)

Alternatively one could add to the resize handler of revDictionary

```
on resizeStack
   set the rect of widget "Dictionary" of me to the contentRect of me
   -- changed
   set the rect of fld "SearchKeyWords" of me to 300,1, the right of me - 8, the top of widget 1
   -- added
   set the textSize of fld "SearchKeyWords" of me to round(the height of field "SearchKeyWords" of me / 1.7)
   -- pass resizeStack
end resizeStack
```

PaulMcClernan:
> I think that field "SearchKeyWords" at the top of the dictionary is redundant since there is another search field for the dicts. And it is awkward to use. (you have to tab into it, one can not click into it to edit)

> Alternatively one could add to the resize handler of revDictionary

> on resizeStack
>    set the rect of widget "Dictionary" of me to the contentRect of me
>    -- changed
>    set the rect of fld "SearchKeyWords" of me to 300,1, the right of me - 8, the top of widget 1
>    -- added
>    set the textSize of fld "SearchKeyWords" of me to round(the height of field "SearchKeyWords" of me / 1.7)
>    -- pass resizeStack
> end resizeStack
> ```
Hi Bernd, good to see you on the HXT train!

Yes that search field is entirely redundant, it runs javascript with the Dictionary's DOM objects to remove the internal search field (Since keyboard input there doesn't work on Linux). In the end that was only meant to be displayed on Linux platform. It was written as a work-around for keyboard focus problems with browser widgets that afflicted it on most Linux distros. 
That's why I commented those lines out in the revDictionary stack I uploaded.

Issue #229, opened 2026-05-18
Strange cursor behavior on Tahoe 26.x.x


RolfKocherhans:
Found a rather strange behavior while trying to set a different cursor.

At first it simply refused to work. After digging through a few older stacks, I finally found one where it did work. Comparing the old and new stacks, I noticed the old one had the “resize stack” property set to false.

That turned out to be the key difference — the issue seems to be related to the resize stack property.
To make things even more entertaining, I then tested on an older macOS VM (Ventura 13.x.x), and there it worked fine in both configurations.

Bernd also reports that on Sequoia 15.x.x it works regardless of the setting.

Could you please check this in your environment? I’m still not entirely sure whether the problem is hiding somewhere else, or if Tahoe 26.x.x is the real culprit here 🤪

I attached a small test stack:

[Cursor Mystery.hyperxtalk.zip](https://github.com/user-attachments/files/27957296/Cursor.Mystery.hyperxtalk.zip)

emily-elizabeth:
can confirm this issue on macOS 26.4.1

emily-elizabeth:
Fixed, just need someone to review it
https://github.com/emily-elizabeth/HyperXTalk/pull/238

emily-elizabeth:
fixed for 0.9.14

PR #238, opened 2026-05-19
Feature/cursor behaviour

emily-elizabeth:
https://github.com/emily-elizabeth/HyperXTalk/issues/229

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,4 @@ build/engine/Debug/Externals/CEF
 build/engine/Release/Externals/CEF
 /thirdparty/libcef
 /Installer/output
+/ide/Documentation/html_viewer/resources/data/api
--- a/engine/src/mac-core.mm
+++ b/engine/src/mac-core.mm
@@ -2003,33 +2003,15 @@ void MCMacPlatformHandleMouseCursorChange(MCPlatformWindowRef p_window)
     MCMacPlatformWindow *t_window;
     t_window = (MCMacPlatformWindow *)p_window;
     
-    // If we are on Lion+ then check to see if the mouse location is outside
-    // of any of the system tracking rects (used for resizing etc.)
-    extern uint4 MCmajorosversion;
-    if (MCmajorosversion >= MCOSVersionMake(10,7,0))
-    {
-        // MW-2014-06-11: [[ Bug 12437 ]] Make sure we only check tracking rectangles if we have
-        //   a resizable frame.
-        bool t_is_resizable;
-        MCPlatformGetWindowProperty(p_window, kMCPlatformWindowPropertyHasSizeWidget, kMCPlatformPropertyTypeBool, &t_is_resizable);
-        
-        if (t_is_resizable)
-        {
-            NSArray *t_tracking_areas;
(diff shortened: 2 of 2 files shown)


BerndN:
Elisabeth, sorry I do not speak 'engine'. Just livecode language. :)

emily-elizabeth:
> Elisabeth, sorry I do not speak 'engine'. Just livecode language. :)

ahh, sorry about that. Can I compile a version for you to test against?

BerndN:
> Can I compile a version for you to test against?

Sorry, I can not test because I am still running Sequoia where the error did not happen

emily-elizabeth:
> Sorry, I can not test because I am still running Sequoia where the error did not happen

Can you make sure that it still works on Sequoia and that I didn't break anything

BerndN:
I just tested with your version pre 0.9.14 and it works on Sequoia as before. I.e. it is not dependent on the stack being resizable or not.

BerndN:
I tested with Rolfs stack. It works but when I open a stack from recent the cursor changes to watch.
However in Rolfs stack using 

```
on mouseLeave
   set the lockcursor to false

```
instead of 
```
on mouseLeave
   set the cursor to arrow
```

Then no more watch cursor.
So the lockCursor should be set to false when done with setting the cursor.

emily-elizabeth:
So is that something with his stack or with the engine?

emily-elizabeth:
Emailed Rolf to take a look at this

BerndN:
> So is that something with his stack or with the engine?

Well It might have something to do with the engine since testing Rolf's stack using LC 10.0.4 it behaves normally without unlocking the cursor. But I am guessing here. What do I know about the engine...

Further testing confirmed my previous observation: using Rolf's stack which does not set the lockCursor to false produces the arrow cursor (which is set on mouseLeave in Rolf's script unless you tools stack to toggle Browse and Edit mode. 

If I test from the message box in 0.9.14 for 
`put the lockCursor`
it gives true.

Doing the same in LC 10.0.4 results in "false"

emily-elizabeth:
> Well It might have something to do with the engine since testing Rolf's stack using LC 10.0.4 it behaves normally without unlocking the cursor. But I am guessing here. What do I know about the engine...
> 
> Further testing confirmed my previous observation: using Rolf's stack which does not set the lockCursor to false produces the arrow cursor (which is set on mouseLeave in Rolf's script unless you tools stack to toggle Browse and Edit mode.
> 
> If I test from the message box in 0.9.14 for `put the lockCursor` it gives true.
> 
> Doing the same in LC 10.0.4 results in "false"

From what I can tell, based on the documentation, because the lockCursor was set to TRUE and never unlocked, it should be returning TRUE in the message box. To me this looks like an error with LC 10.0.4

PaulMcClernan:
FYI, If you want to add some open source cursors (like a 'Pen tool' inkwell pen cursor for example ;) ), I made a stack that can import XCur commonly used on *Nix platforms: https://www.openxtalk.org/forum/viewtopic.php?p=16498&hilit=xCurs#p16498

PR #239, opened 2026-05-19
fix(engine/linux): honour system natural-scroll preference for wheel …

emily-elizabeth:
…events

Commit c036cbcd7 changed the GDK scroll-direction mapping to match the non-natural-scrolling convention (GDK_SCROLL_UP → content scrolls up). That is correct when natural scrolling is off, but broke the behaviour for users who have natural scrolling enabled in GNOME System Settings: HxT was scrolling in the opposite direction from every other application on their desktop.

MCS_getnaturalscrolling() now has a Linux implementation that queries
  org.gnome.desktop.peripherals.mouse natural-scroll
via the gsettings(1) command-line tool (forked once at first call, then cached in a static).  On KDE, Xfce, or any desktop where the schema is absent, the command produces no output and the function returns false, preserving the existing behaviour.

In lnxdclnx.cpp the GDK_SCROLL handler negates both deltas after the direction switch when MCS_getnaturalscrolling() is true, so the sign conventions line up with what every other GTK application delivers.

### DIFF
--- a/engine/src/lnxdclnx.cpp
+++ b/engine/src/lnxdclnx.cpp
@@ -586,15 +586,19 @@ Boolean MCScreenDC::handle(Boolean dispatch, Boolean anyevent, Boolean& abort, B
                         // Is this a mouse scroll event?
                         if (MCmousestackptr && t_event->type == GDK_SCROLL)
                         {
-                            // GDK direction names reflect the natural-scrolling convention;
-                            // map to signed deltas matching the scrollWheel message contract.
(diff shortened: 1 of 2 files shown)


mwieder:
No... still the same

emily-elizabeth:
Claude reverted the changes, should be good to go now.

mwieder (approved):
Looks good. Sorry for the delay in reviewing.

emily-elizabeth:
> Looks good. Sorry for the delay in reviewing.

Take your time, we're not going anywhere :)

PR #240, opened 2026-05-19
Update revDictionary.rev

emily-elizabeth:
The Guides section needs to be cleaned up of mention of Livecode and the screenshots need to be removed.


PaulMcClernan:
If you look at those docs/Guides/ files that I pushed  (from OXT DPE) , those are already pretty-much completely de-branded (as far as I know), **even the screenshots** used in Guides were modified with an image editor to remove and LC references. 

The Builder stack using the revDocsParser library can rebuild the Docs sqlite database and JSON/JS/HTML versions from those .lcdoc files. The output from doing that with these debranded files looks like what I put online here: https://openxtalk-org.github.io/OpenXTalk_Syntax_Dictionary/

FYI, If you change the API categories in the .sqlite db then you must also change them in the revIDELibrary as that's where '**LiveCode** Script" display names in that drop-down menu comes from, and check revDocsParser as that may need to be updated to match any changes as well. 

The docs system is a bit convoluted in my opinion. That combined with the fact that the Browser Widget has keyboard input-focus problems on most Linuxes (which hopefully will be fixed) led to OXT Lite re-implementing it completely differently, as plain-text based system for docs (.lcdoc were stripped of markdown tags to plain .txt files).

PaulMcClernan:
Ah "'click at' doesn't work"?  Since when? 
Well OK, I commented that line. Will push in a few. 
Anyway, that was part of attempts at work-arounds for Linux keyboard input-focus bug with Browser Widget. You see the input focus would sometimes work when the page is first loaded into the Browser Widget, so I was forcing page reload to see if I could force keyboard input focus (hopefully GTK3+ support and a new WebkitGTK based browser widget will resolve these issues 💯 )

emily-elizabeth:
"click at" works for clicking at a location, not at specific line chunk.

emily-elizabeth:
This is what I see when I look at the Guides section

image

bwmilby:
Do we really want the dictionary and guides as two tabs of one window?  Any consideration of just making the guides their own window?

emily-elizabeth:
I think the guide should be a separate window.

emily-elizabeth:
Okay, I just spent the last couple hours restoring the documentation. Please leave it the way it is. Guides can be it's own window.

PaulMcClernan:
OK, Oh well, I just spent hours trying to fix up conflicts. I hadn't realize you had already been tinkering about in files like tools_builder.livecodescript (which I restored to what you had). 
Are you using the Builder stack and the lcdocs to rebuild the dictionary data files? 

I agree that guides _should_ have their own stack, but opted for shorter road of keeping tabs as inherited because it was the much easier route with minimal edits to the various IDE library scripts that tie in to Docs.

emily-elizabeth:
All the old mobile docs were restored, the HTML wasn't reading the docs for the script libraries, there was documentation entries for externals we do not have and the documentation cache wasn't being populated fully. So for now, lets just leave the documentation alone and work on the Guides, which can be opened from the Help menu

PaulMcClernan:
That's due to the debranding of the API labels. If you want to get rid of "LiveCode" from the Dictionary, it's a necessity to altering rebuild  all of that, as well as tie-ins in IDE libraries specifically revIDELibrary and revDocsParser

PaulMcClernan:
Of course any Externals or Mobile stuff can be removed by deleting the .lcdoc for it and then rebuild the dictionary data files (also need to flush the docs cache each time)

emily-elizabeth:
> Of course any Externals or Mobile stuff can be removed by deleting the .lcdoc for it and then rebuild the dictionary data files (also need to flush the docs cache each time)

I had did all of that before.

PaulMcClernan:
AH OK Sorry, I just dumped over the Docs folder from OXT DPE (where I've tried to maintain support for everything that was Inherited by Community)

PaulMcClernan:
But the other modifications I'm talking about are important to consider if you want to change this:
Screenshot 2026-05-21 at 1 26 30 PM

PaulMcClernan:
And I updated the Guide files so they look like this:
Screenshot 2026-05-21 at 1 26 48 PM

emily-elizabeth:
> AH OK Sorry, I just dumped over the Docs folder from OXT DPE (where I've tried to maintain support for everything that was Inherited by Community)

No worries, I'm just lazy 😉

PaulMcClernan:
That's the nice about version control, can always revert fairly easily (I've got like 10 versions of the IDE on my drive, lol)

PaulMcClernan:
I don't want to step on any toes here, it's just that I don't like watching a _THIRD_ Community fork redo any editing that I already did 3-4 years ago :-)

PR #245, opened 2026-05-20
engine: NOOP create toolbar if named toolbar already exists

emily-elizabeth:
Calling `create toolbar "mainBar" in this stack` a second time was re-cloning the template toolbar over the existing one, wiping all previously added items and leaving the toolbar in an unrecoverable state.

MCInterfaceExecCreateControl now checks the target stack's control list before cloning: if a CT_TOOLBAR object with the same name already exists, the command returns its long ID in `it` and exits without creating anything. The check is wrapped in #ifndef _SERVER to mirror the existing server-build guard on toolbar support.

Also removed a leftover debug fprintf/fflush from
MCToolbar::SetItemLabel in exec-interface-toolbar.cpp.

### DIFF
--- a/engine/src/exec-interface-toolbar.cpp
+++ b/engine/src/exec-interface-toolbar.cpp
@@ -168,9 +168,6 @@ void MCToolbar::GetItemLabel(MCExecContext& ctxt, MCNameRef p_item,
 void MCToolbar::SetItemLabel(MCExecContext& ctxt, MCNameRef p_item,
                              MCStringRef p_label)
 {
-    fprintf(stderr, "[SetItemLabel] called on toolbar %p item=%p\n",
-            (void*)this, (void*)p_item);
-    fflush(stderr);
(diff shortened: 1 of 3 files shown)


mwieder:
Well, OK - it *doesn't* create a second toolbar if one already exists. So there's that. But closing the toolbar test stack after clicking the 'create toolbar' button quits the IDE dramatically with a signal 11.

emily-elizabeth:
Verified on macOS and Windows

emily-elizabeth:
I'm not even getting a toolbar created, but I am getting the crash ;)

mwieder:
Welcome to my world. But I thought the whole idea was that a toolbar was *not* supposed to be created if there was already one in existence.

emily-elizabeth:
I'm not even getting the first toolbar to show up.

mwieder:
In the project browser you don't see a toolbar widget?

PR #252, opened 2026-05-21
Mostly De-Branding stuff

PaulMcClernan:
ALSO Converted Debugger Library in-line documentation into lcdoc markdown so it can be parsed by revDocsParser at some point (could be helpful for IDE development).


bwmilby:
I need to check once I get to a computer but I don’t think you need to strip out the docs.  I’m pretty sure that the script can be parsed directly to extract the docs.  The dev builds always had a third dropdown for IDE docs that were not present in the release versions.

PaulMcClernan:
> I need to check once I get to a computer but I don’t think you need to strip out the docs. I’m pretty sure that the script can be parsed directly to extract the docs. 

I didn't 'strip out the docs' or any of the pre-Browser Widget Dictionary in-line documentation, what I did was I changed it to comply with the format that .lcdoc markdown / revDocsParser expects. If you open it using the Extension Builder stack and use it's package button, it will parse either an lcb or a .livecodescript file for lcdoc formatted in-line markdown and output a API.lcdoc file which can then be added to the dictionary 'source' folders.

>The dev builds always had a third dropdown for IDE docs that were not present in the release versions.

I'm aware of the 'IDE' catagory/label, but didn't realize it was previously only added in for 'Development' builds (but I guess that makes sense if the category is only going to hold IDE related docs). I'm not exactly sure how to tag them with the lcdoc markdown so that they wind up under that API category though (is there an 'API' tag like 'API: IDE'?).

PaulMcClernan:
Just to elaborate, the debugger library had in-line docs already but they where written like this:
```
# Parameters
# pContext : a debug context in the format returned by revDebuggerContexts. Note this includes a number as the first item.
#Description
#
``` 
etc.

but revDocsParser expects this format:
```

/**
Parameters:
 pContext : a debug context in the format returned by revDebuggerContexts. Note this includes a number as the first item.
Description:
 Sets the debugger context to pContext. This value will only remain while the debugger remains on its current line when the line changes (e.g. because the user has pressed "Step over") the context will be reset to the current one.
*/
```
NOTE: the **double** asterix  at the start (seems to be needed to indicate its an lcdoc markdown in-line doc and not a regular multi-line comment)

bwmilby:
Ok, I understand now.  I made an assumption (bad of me) that "convert" equated to "extract".  It's been 8 years since I've done much with documentation, so I'll need to refresh my knowledge and then dig around to figure out the answer to your question.  I believe that at one point there was a desire to fully document everything in the IDE but until enough was done it wasn't going to be present in builds that were released.  Personally I believe that if the docs are written then they should be published for view.

PaulMcClernan:
> Personally I believe that if the docs are written then they should be published for view.

Same here, it's nice to have IDE library reference if you're working on things for the IDE (including add-ons) that's why I've done some of this updating and extracting to add it to the Dictionary, but I do think it should be in its own category since it contains only things that apply to IDE development.
Another option might be to keep a separate IDE Developer build of the Dictionary, and have it viewable online like here:
https://openxtalk-org.github.io/OpenXTalk_Syntax_Dictionary/

PaulMcClernan:
Just checked upstream and the duplicates are still in there. I'll have to fix that later tonight when I get back to my dev machine.

PR #254, opened 2026-05-21
Fix/vlc audio only playback

emily-elizabeth:
Audio only players would not play due to a height check, so the height check is only used when the media has a video track. Fixed for macOS and Windows. Linux was not affected.

### DIFF
--- a/engine/src/vlc-player.cpp
+++ b/engine/src/vlc-player.cpp
@@ -2446,12 +2446,47 @@ void MCVLCPlayer::Start(double rate)
     // applies the deferred rect and calls [view setFrame: actualRect], the
     // callback fires (async, to avoid re-entering AppKit's layout pass) and
     // starts VLC at that safe moment.
+    //
+    // Audio-only exception: for media with no video tracks (e.g. MP3), the
+    // player's video area has zero height — doSetGeometry is always called with
+    // a zero-height rect, so the NSView frame is perpetually zero and the
+    // frame-ready callback never fires.  Audio output in VLC uses a completely
+    // separate pipeline that doesn't need an NSObject view, so we can skip the
+    // deferral and play directly.
     if (!m_offscreen && m_view != nullptr)
     {
+        // Detect audio-only media: check parsed track list for any video track.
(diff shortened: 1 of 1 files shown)


PaulMcClernan:
libVLC can use FluidSynth to playback MIDI files, so I'm wondering are there any plans to add VLC extra things like that? On macOS libFluidSynth is not needed for MIDI playback because it has AVMIDIPlayer API (already have an xBuilder wrapper around that). 

Come to think of it on macOS libVLC is not needed at all for a most of today's common media format because AVFoundation comes with macOS. Likewise MS Windows has it's own Media Foundation (including MIDI playback), and there's GStreamer framework on Linux (which I think depends on some of the same libraries as libVLC). 

But I suppose using the same media library across platforms would make maintaining it easier.

emily-elizabeth:
I had looked at using the system media playback features, but they were too limited by platform, so I went with VLC to give a good media playback feature for HyperXTalk that would be cross-platform.

PaulMcClernan:
Resynced my fork, re-compiled everything from fresh re-download (backing off from any Dictionary editing for now).
No problem compiling. VLC already installed (I've used VLC for like 25 years now). Tried to make a player control instance on a card crashes engine on MacOS 15.x

PaulMcClernan:
My 2cents... Personally I think keeping to using AVMedia framework on macOS makes the most sense, it already works. I think Windows build was using older media API so that might need work, but the macOS player is fairly solid and doesn't require any additional libraries. Linux was the one in the worse shape as far as media player, it wouldn't even show any GUI controls. Tom Perry was looking to switch Linux player to GStreamer but don't recall if he finished the work.

What limits are you looking to not be avoid?

emily-elizabeth:
> Tried to make a player control instance on a card crashes engine on MacOS 15.x

Did you do the last step `make package-mac-bin`? I found without doing that it crashes, but it could also be a macOS 15 thing. Did you get a crash log?

emily-elizabeth:
New PR https://github.com/emily-elizabeth/HyperXTalk/pull/346

PR #255, opened 2026-05-22
Scriptified the Answer Dialog

mwieder:
Left the file extension at livecodescript
because hyperxtalkscript files aren't yet loaded into the IDE system


emily-elizabeth (commented):
Can you update line 99 to read
`put "210093 210092 210094 210097" into tIcons`

PR #257, opened 2026-05-22
Feature/dark mode aware color system

emily-elizabeth:
**Add semantic color system** (com.hyperxtalk.library.appearance)

Adds a new script library that gives HyperXTalk developers a dark mode-aware color system at the script level, without requiring a systemAppearanceChanged handler in every stack.

**How it works**

Developers define named color roles — each carrying a light and dark variant — and annotate their controls with cSemantic* custom properties. The library holds a single systemAppearanceChanged handler in the back-script chain that walks all open developer stacks and applies the correct color automatically when the OS appearance changes.

**Public API**

- defineSemanticColor pName, pLightColor, pDarkColor — register or override a color role
- semanticColor(pName) — query the current-appearance color for use in custom drawing code
- applySemanticColors [pTarget] — force an immediate re-apply, e.g. on openCard

**Supported custom properties → LiveCode properties**

Custom property -> HyperXTalk property
cSemanticFillColor -> fillColor
cSemanticForegroundColor -> foregroundColor
cSemanticBackgroundColor -> backgroundColor
cSemanticBorderColor -> borderColor
cSemanticShadowColor -> shadowColor
cSemanticHiliteColor -> hiliteColor

A default nine-role palette ships out of the box: background, surface, label, label_secondary, accent, accent_label, separator, input_background, and destructive. All roles can be overridden or extended.

In development mode, IDE stacks are skipped so revpalettebehavior continues to own the IDE's own appearance. In standalones the filter is a no-op.

**Files changed**

extensions/script-libraries/appearance/appearance.livecodescript — the library
extensions/script-libraries/appearance/api.lcdoc — documentation
extensions/extensions.gyp — registered in the packaged extensions build

### DIFF
--- a/extensions/extensions.gyp
+++ b/extensions/extensions.gyp
@@ -85,6 +85,7 @@
 				'script-libraries/httpd/httpd.livecodescript',
 				'script-libraries/qr/qr.livecodescript',
 				'script-libraries/shakecontrol/shakecontrol.livecodescript',
+				'script-libraries/appearance/appearance.livecodescript',
 			],
 			
 			'dependencies':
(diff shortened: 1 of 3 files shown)


emily-elizabeth:
Made this a package manager download

PR #261, opened 2026-05-22
fix(linux): resolve player controls not responding

### LINKED ISSUE #183: [linux] the player object doesn’t respond to stop or pause commands.
the player object doesn’t respond to stop or pause commands.

docmeth02:
Resolves #183

### Description
This PR resolves the issue where the `player` object does not respond to `start`, `stop`, or `pause` commands on Linux.

#### Core Findings
1. **Active Architecture**: Modern Linux desktop builds of HyperXTalk compile and use the modern cross-platform libVLC-based player backend (`engine/src/vlc-player.cpp` via `engine/src/player-platform.cpp`). The legacy `engine/src/lnxmplayer.cpp` (which spawned `mplayer`) is completely compiled out when `FEATURE_PLATFORM_PLAYER` is set.
2. **Root Cause**: On headless/CI environments or clean Linux environments where only the base development dependencies (`libvlc-dev` or `libvlc5`) are installed without VLC plugins, the `libvlc_new()` call fails to initialize the engine backend. Because of this, player creation and commands fail silently.
3. **Resolution**: Installing the standard VLC plugins packages (`vlc-plugin-base` and `vlc-plugin-video-output`) restores full libVLC initialization. Once these plugins are present, all player control commands (`start`, `stop`, `set the paused to true/false`) function flawlessly out of the box.

#### Changes Introduced
1. **Automated Test Suite**: Added `tests/lcs/core/multimedia/player_control.livecodescript` along with a minimal audio asset (`tests/lcs/core/multimedia/silence.wav`) to thoroughly verify player creation, filename loading, properties (playLoudness, looping), and control commands (start, pause, stop) on Linux and other platforms under Xvfb.
2. **Build System Fix**: Retained the output directory creation fix in `Makefile.linux` to ensure Release compilations do not fail during extension packaging.

### Verification
- Ran the new player control test suite inside Xvfb:
  ```bash
  xvfb-run ./linux-x86_64-bin/standalone-community ./tests/_testrunner.livecodescript run ./tests/lcs/core/multimedia/player_control.livecodescript
  ```
- Result: **All 11 assertions passed successfully** (`PASS:  ./tests/lcs/core/multimedia/player_control: TestPlayerProperties [11/11]`).
- Ran all 28 multimedia tests:
  ```bash
  xvfb-run ./linux-x86_64-bin/standalone-community ./tests/_testrunner.livecodescript run ./tests/lcs/core/multimedia
  ```
- Result: **All 28 tests passed successfully**.


mwieder:
Awesome. Yes, after installing the two vlc plugin packages this works flawlessly. Standalones also are built properly with the vlc libraries included.

docmeth02:
thanks for testing @mwieder 
ready to merge then @emily-elizabeth

emily-elizabeth:
thanks for testing @mwieder
thanks for your hard work @docmeth02

mwieder:
So now that we've gotten this far (this seems like maybe an appropriate place for this comment) there are a couple of things I'd like to change:
1. Once a video/audio file has been played, issuing a 'play' command again has no effect.
Note: I do see Start: re-set xwindow=86008449 in the terminal
3. The 'stop' command acts like a 'pause', toggling the play state.

I'd like to have a 'pause' command that replaces the functionality of the current 'stop' command (easy fix).
Then I'd like to have the 'stop' command actually stop the player so that subsequent 'play' commands would start over. And this probably means resetting the state of the player object as well.

I also note that the 'loop' option has no effect.

Comments?

emily-elizabeth:
I don't think changing the "stop" to restart the video makes sense. If you press stop on a VCR/DVD Player, the movie doesn't start over.

mwieder:
True, but there *is* a way to start over on a VCR. And continually pressing 'stop' doesn't toggle it. I just don't think the current set of commands is intuitive.

emily-elizabeth:
I've never used the player control, so not too certain what the commands are for it. Maybe there needs to be a command to restart the video or can you scrub (in code) to frame 1?

mwieder:
Here's a test file with buttons that act in what I think is a more intuitive approach. Luckily everything I want from this can be accomplished at the script level.
[playtest.hyperxtalk.zip](https://github.com/user-attachments/files/28195922/playtest.hyperxtalk.zip)

PR #262, opened 2026-05-23
[[Issue 249]] Update doc/guide build process

### LINKED ISSUE #249: Documentation Build Issue
In the original project, the `.lcdoc` files were processed prior to deployment of a new build.  Most of the files were located under `/docs/*` of the repo.  Due to the IDE being a submodule, there were a few nested under `ide/Documentation/*`.  This morning the updates to the project duplicated quite a few files from the engine side to the IDE side which should not be needed.

On the first launch of a newly built project it would build the dictionary from the raw files using repo paths.  `api.sqlite` really should not be saved in the repo as it is a derived file much like the build files for all of the `LCB` code.

The build process currently copies the raw files into the application which should not be necessary.  Since I have worked with the doc system in the past, I'd like to propose that I surface the build command to properly build the docs for the standalone and include that in the build process.

This project may be useful (I have not updated it/ensured it works with HXT yet):
https://github.com/bwmilby/DocEditorPlus

It may be as simple as launching HyperXTalk from `mac-bin` before the package step.  This will have the IDE build the documentation `.js` and `.sqlite` files.  I just need to do some additional testing to see if the old files need to be cleared first (I suspect this is the case).

bwmilby:
This update streamlines the build process so that the documentation can be built reliably prior to creating the application package for distribution.  Doing it this way saves time on first launch for the end user since the majority of the docs are already compiled.  The guide file is built here, but restoring display in the IDE will be handled in a different PR.

After compile, the next step to build the docs is now shown: `make docs-mac`.  Optionally, the docs can be built by launching HyperXTalk from the mac-bin folder.  To ensure the docs are freshly built, the previous cache is removed at the end of the compile step.

- Update the makefile (mac) to not copy docs
- Add some additional echo statements (begin/end/next step)
- Clear document build cache at end of compile (`ide/Documentation/html_viewer/resources/data/*`)
- Update `.gitignore` files to exclude `api.sqlite` and extension `api.lcdoc` files
- Untrack `api.sqlite` file
- Update args for wkhtmltopdf so that the latest available version works (must explicitly allow file access)
- db driver path needed to be set for builder_tool to use databases with `-ui` flag
- docs_builder needed to be branded to LiveCode due to file paths (eventually needs to move to HyperXTalk, removing branding totally will be more difficult)
- added command to move built guide files to docs-build folder (this does break the relative html references but avoids the files accidentally being committed to the repo)
- removed stale notes/user guide pdf files (old version, blank document)
- correct `build-docs` in makefile to reference proper targets
- added instruction to `make docs-mac` next after compile
- added target for `docs-mac` that lists the next: `make package-mac-bin`
- correct parsing for parameters named `tag`

Closes #249 

[It saves about 4.5MB in the app package and removes over a thousand files]


bwmilby:
`make build-docs-guide` can now be used to generate the 490 page guide (after the duplicates are removed).  It does require `wkhtmltopdf` which isn't available compiled for arm64.  I was able to get it to work with Rosetta.  URL with the instructions I used is in the makefile.  `build-docs-api` target is there but I get a database error.  `build-notes` will not complete due to `mergMarkdown` not being available to download.

bwmilby:
For the purposes of this PR, I've found a suitable replacement for `mergmarkdown`:
https://github.com/commonmark/cmark
It can be installed via brew on Mac.  While an external may be nice in the future, this will allow progress on getting the notes created again.

emily-elizabeth:
Considering we aren't doing the build notes like LC used to do, is this really needed?

bwmilby:
The notes piece specifically may no longer be applicable, but it is a minor piece of the overall PR.  I think that it may be useful to have a discussion on the overall release notes process.  If we want to scrap it, there would be quite a bit that could be removed from the repository.  I do think the guide would be good to bring back though.  Only need to generate that on one platform and upload the PDF artifact.

emily-elizabeth:
Well with each version, there is the release notes posted with it. We don't need to copy what LiveCode used to do - I'd say scrap it. The guide, is that the thing that we're going to add back in to the IDE, that used to be part of the documentation (dictionary) window?

bwmilby:
Correct, the guide is implemented as a html site that was shown in the same window as the dictionary but a different tab.  The IDE also had a link to a PDF version of the guide.  I would suggest that we make the PDF available for download separately (~15mb) and get the guide back by using a browser widget window.

If you would like, I can add a commit to rip out the whole notes process.  Could easily be done later as a separate PR too.

emily-elizabeth:
Yeah, the guide is going to be in its own window, not merged with the dictionary like before. Yes, the Guide PDF file can be placed on the hyperxtalk.com website for download.

Do it as a separate PR so we keep things tidy (something I'm guilty of not doing 😉 )

bwmilby:
I just noticed that the original .ignore had the guide build files ignored.  I had coded to move them after build, but probably better to leave them and add to the ignore.
I also just realized that GitHub Desktop allow easy combining of commits which will allow a much cleaner merge since back and forth types of commits can be eliminated.  It does end up requiring a force push though.

bwmilby:
This one should be ready to merge.

emily-elizabeth (approved):

emily-elizabeth:
> - Update the makefile (mac) to not copy docs, add some additional echo statements, launch HyperXTalk after compile to build docs

So what will I need to do? This is my current build steps

```
make config-mac
make prebuilt-mac
make bump-revision (only do this when on main)
make compile-mac
make package-mac-bin
```

bwmilby:
before package, you will need to do `make docs-mac` which will build all of the documentation fresh (compile will clean out the folder).  This eliminates the need to have any of the doc files stored in the repo.  For the other OSs, can do `make build-docs-api` (which is what the Mac version calls.  the Mac version just adds some additional instructions)

One thing I did with the makefile is to include the next step required for the user.  Config looks for a file ("prebuilt/lib/mac/libmysql.a") and will advise to do prebuilt if missing or compile if present.

PR #266, opened 2026-05-23
Fix Find & Replace highlight unreadable in Dark Mode

### LINKED ISSUE #264: Find and Replace text use a green highlight color
The Find and Replace text currently use a green highlight color for found text. In Dark Mode, the text underneath can be difficult to read.

Image

Would it be possible to add a preference for this, perhaps something like:

“Found text selection color:”

This would allow users to choose a highlight color with better contrast for different themes.

Suggestion:

Image

emily-elizabeth:
Add editor,darkFindColor preference (default: muted steel blue 38,79,120) as a dark-mode counterpart to editor,findcolor (existing green 0,200,0). findMarkResult now picks between the two based on the systemAppearance, so light-coloured syntax tokens from dark-mode colour schemes remain legible against the find highlight background.

Both keys are user-configurable via sePrefSet; Preferences dialog UI to follow as a separate task.


emily-elizabeth:
verified on macOS and Windows

PR #267, opened 2026-05-24
Scriptified the Ask Dialog

mwieder:
The Ask Dialog is now a script-only stack.
Unfortunately I haven't been able to incorporate Mikey's contribution of selecting the text of the input field.
Or rather I can do it, but then it ignores returns and backspaces and... until you click on it.
That seems more confusing, and maybe there'll be a way around this later on now that it's just text.


emily-elizabeth (commented):
Can you update line 87 to read
`constant kMacIcons="210093 210092 210094 210097"
`

emily-elizabeth:
oh, looks like I could do it myself, that's cool 😄

PR #269, opened 2026-05-24
Cleaned things up a bit - minor edits

mwieder:
Nothing major here. Mostly just cleaning things up.
Added a note as to why I punted on auto-selecting the input text.
Maybe a fix or a workaround will pop up in the future.


emily-elizabeth (approved):
Good to go

PR #272, opened 2026-05-25
Reorganize execution error codes in executionerrors.h

bwmilby:
executionerror.h additions need to follow a specific format.  They are added to the end of the file and each one gets the next sequential number in series `{EE-9026}` to `{EE-9028}` in this case.

Error identified while working on the test suite.


mwieder (commented):
Looks good.
I've never liked this approach to the error messaging. It's awkward and error-prone. At some point I'd really like to do a complete makeover here and use enums instead of relying on comments for actual data.

emily-elizabeth (approved):

PR #278, opened 2026-05-26
removed hardcoded TRUE from DevPreferences to allow the toggling of the strict compilation mode

### LINKED ISSUE #274: cannot disable Strict Compilation Mode
cannot disable Strict Compilation Mode

emily-elizabeth:
removed hardcoded TRUE from DevPreferences to allow the toggling of the strict compilation mode


emily-elizabeth:
verified on macOS

emily-elizabeth:
verified on Windows

PR #279, opened 2026-05-26
Changed the default settings when creating a new preference file

### LINKED ISSUE #256: HXT file settings
A very old system for keeping files in memory for opening faster when used again is no long of any value, but is a source of great annoyance when you think you've closed a file but it still lingers in memory.

Image

I would suggest maybe removing it from memory *always* is the way to go (as per my choices above). 
File access is so fast in this day and age that I can't see this is of any benefit, quite the opposite...

emily-elizabeth:
Changed the default settings when creating a new preference file


emily-elizabeth:
Default settings have been changed.
image

emily-elizabeth:
verified on Windows. can someone test on Linux please?

PR #280, opened 2026-05-26
Fix to #276

mwieder:
Both ask and answer dialogs are no longer resizable
Prompt text of answer dialog is no longer automatically copied
(was a request, but I see the problem with it)
Pressing control-C/command-C will copy the prompt text


emily-elizabeth:
Need this fixed. Create a new stack, add a control, close the window. The answer sheet will be a different size and you can't see the buttons.

image

mwieder:
Not sure what happened there... maybe pasted some older test code... seems fixed now after a bit of fiddling.
Interesting that I didn't have problems with it before though.

emily-elizabeth:
Still have the same issue

image

mwieder:
That's bizarre. I'm not seeing that here. And the stack resizing is the same as the original behavior script. Any idea what's going on?

emily-elizabeth:
I've looked through the code and can't see anything that would be causing it.

mwieder:
Is this only on OSX or do you see this on Windows as well?

emily-elizabeth:
In all honesty I haven't tried Windows, give me a few minutes

emily-elizabeth:
Can't test on Windows, it keeps saying you deleted the branch or fork

mwieder:
Eh? Even more bizarre. I *never* touch Windows. Are those maybe the dlls that keep getting regenerated?

emily-elizabeth:
You'll notice that those DLLs haven't cropped up lately. I've learnt a little trick - shove everything into a stash and just ignore them 😃

mwieder:
I think I've been doing that too. Those and that pesky sqlite database.

emily-elizabeth:
When I step through the PreOpenStack handler the dialog box opens with the right height and width, but if I just let it run I get the same as the picture  above

mwieder:
Er... *I* closed this? I thought I just pulled your latest commits.

PR #285, opened 2026-05-28
[[Issue 284]] Remove duplicate documentation files

### LINKED ISSUE #284: Duplicate documentation in repo
Currently there are over 40 duplicate documentation files in the repo.  This results in the generation of a bloated guide with a number of double entries.  This can also lead to duplicate efforts as the incorrect file could be edited or even worse edits could be done to both files causing additional work merging them together.

bwmilby:
Remove duplicate copies of documentation files
Renamed `Extending OXT.md` to HyperXTalk
Created symlinks to files from `specs` folders
Removed obsolete build target files
Correct style guide link in `contributing_to_docs.md`

Closes #284


bwmilby:
This one should be ready to merge.

bwmilby:
Resolved the concern that Mark would have raised.  I restored the Android build file but it will need validation and updating to actually be able to build.

emily-elizabeth:
> Created symlinks to files from specs folders

Will the symlinks work on all platforms (macOS, Windows and Linux)?

bwmilby:
The symlinks are just for the repo.  On platforms that don't support them, it is just a text file that contains the location of the actual file.  Files in the `specs` folder are not copied to the app.  The files were copied to the `guides` folder and I think it makes sense to include them in the app.  Instead of having them in the repo twice, I opted to add a symlink since technically they are specs.  The other way to address would be to includes the specs folder documents with the guides.

emily-elizabeth (approved):

PR #287, opened 2026-05-28
Scriptified answer dialog

mwieder:
This takes care of the Answer dialog with platform-specific decorations.
I'll submit the Ask dialog update as a separate PR to keep them separate.

PR #290, opened 2026-05-29
no more revGoUrl

mwieder:
replaced revGoUrl with launch url


emily-elizabeth (commented):
I don't see the "decorations" fix or the "launch url" fix either

emily-elizabeth:
The changes are on PR 287, do we even need this PR?

mwieder:
I think that's correct. #287 has the latest. Not sure why this ended up as a separate PR.

PR #293, opened 2026-05-30
Delete extra files 052926

bwmilby:
`revsblibrary.livecodescript` was duplicated in the repo again.  There are different changes between the two files, but the one living in `ide-support` is both the correct location and has what it should (it has 2 or 3 additions not in the one being deleted).

removing several files pertianing to targets no longer being used (Android, iOS, emscripten)


bwmilby:
In particular, `private function __HxtpmScriptLibraries` was included in different locations in the two files.

emily-elizabeth (approved):

mwieder:
Actually as long as the android etc libraries in ide-support aren't doing any harm, I've got use for them and would like to leave them in place. If there are no objections. I can siphon them off into a branch of my own if it comes to that, but the deletion and reinsertion work seems unnecessary.

bwmilby:
Not so much harm, but they are being loaded when there is no reason to for this project.  The home stack loads everything in that folder in development mode.  IDE docs are a mess due to how the list is generated, but extra stuff in that folder will make it more difficult (I was going to use that folder as a base instead of the current method of home stack files).  There is a very large amount of stuff that could be purged by eliminating Android/iOS/emscripten code and docs.  It's all going to still be in the original livecode repo too, so not difficult to recover if ever needed.
Having said all that, I'm not that strongly opinionated at this time.

mwieder:
I have in mind bringing back at least Android as a side project in the future once things settle down here, and it would be much easier if some of the mechanism was already in place. So can we maybe just put this aside for the moment and come back to it in a bit?

bwmilby:
Sounds good to me.  There are probably a handful of file deletes in flight but no significant code updates that pull out mobile stuff inside remaining files.  I know the dictionary has been updated quite a bit though.

emily-elizabeth:
Was the Android stuff put back in place for Mark?

bwmilby:
I'll do that this morning.

bwmilby:
Force push restored the Android files that had been initially deleted.

PR #294, opened 2026-05-30
Cleanup/documentation for web

emily-elizabeth:
removed outdated documentation


bwmilby:
I think you may have removed a bit too much.  There are two entries for many things where one is a command and the other is a function.  For example `revdb_commit` is the function version of the `revCommitDatabase` command.  The way they are used is slightly different.  You can use the function directly in an `if` statement instead of having to use the command and then check `the result` separately.  In my opinion, we need to retain docs for both variants.

If you like, I'll go through the PR and generate a reverse for the files that I believe should be retained, but don't want to duplicate your efforts.

emily-elizabeth:
I thought `revdb_commit` was the older method of calling. I think having two different methods to do the same thing could be confusing, but what do I know

bwmilby:
While the effect is the same (db commit), they are different.  Functions return values directly.  Commands can have a return value, but it must be accessed differently.  Both were introduced in 1.0.0.

emily-elizabeth:
Learn something new every day. I guess put them back in then.

bwmilby:
Just realized that you had reset "introduced" to 1.0.0 for everything.  To correct myself, the function was introduced in 1.1 and the command added at 2.0 but I still think both are useful.  I'm guessing 2.0 was also when they renamed `revdb_connect` to `revOpenDatabase` but since they are both functions Livecode just renamed it in the dictionary and added a synonym to the original name.

emily-elizabeth:
yeah, I think the `revdb_` commands are all the original way they named them and then they renamed them to the ones that I left. The commands are still in the engine/external so they can still be used, just the documentation is gone so people use the new commands

bwmilby:
Actually all of the `revdb_` entries are functions.  Commands were added later.  Functions and commands are distinct.  You can actually have a function and command with the same name in a script.  I'm pretty sure the originals were never deprecated on purpose since they are useful.  This is probably something that really should have more input before eliminating documentation on things still in the engine.

emily-elizabeth:
If you feel they should still be in the documentation by all means feel free to put them back in I defer to your knowledge on the matter 😊

bwmilby:
I created a PR to restore the revdb functions to this branch.

PR #297, opened 2026-05-30
[lin] engine: add Linux toolbar backend with GDK/Cairo/Pango

emily-elizabeth:
Implements MCToolbarLinuxBackend, the Linux platform backend for MCToolbar. The toolbar is rendered as a child GdkWindow positioned at the top of the stack window, drawn entirely with Cairo and Pango. This is architecturally consistent with the rest of the Linux engine, which uses raw GdkWindow* throughout rather than GtkWidgets. engine/src/lnx-toolbar.cpp (new)

Creates a child GdkWindow matching the parent window's visual and colormap, avoiding compositor transparency issues on RGBA-visual stacks Draws a gradient background, separator lines, icon (GdkPixbuf, scaled), and label per item using Cairo Uses pango_cairo_create_layout / pango_cairo_show_layout for label text so Unicode button labels render correctly Registers Xlib event filters for expose (redraw), button press (hit-test → itemClicked), and motion/leave (hover highlight); a second filter on the parent window handles ConfigureNotify for resize Create() is idempotent — calls Destroy() on entry so a second createToolbar without a prior close reinitialises cleanly rather than leaking the old window and double-registering filters Three-pass layout: fixed widths → flex-space distribution → x-position assignment

engine/src/linux.stubs

Added GDK: gdk_window_add_filter, gdk_window_remove_filter, gdk_drawable_get_visual
Added Cairo: cairo_pattern_create_linear, cairo_pattern_add_color_stop_rgb, cairo_set_source, cairo_pattern_destroy, cairo_save, cairo_restore, cairo_translate, cairo_scale, cairo_paint_with_alpha, cairo_set_source_rgb, cairo_set_line_width, cairo_move_to, cairo_line_to, cairo_stroke, cairo_select_font_face, cairo_set_font_size, cairo_text_extents, cairo_show_text; corrected argument types on cairo_set_source_rgba and cairo_rectangle
Added pangocairo: pango_cairo_create_layout, pango_cairo_update_layout, pango_cairo_show_layout
Added pango: pango_layout_get_pixel_size

engine/src/exec-interface.cpp

MCInterfaceExecCreateControl: when p_type == CT_TOOLBAR, walk the stack's controls list before creating anything; if a toolbar already exists, set it to its long ID and return — matching the macOS and Windows no-op behaviour

### DIFF
--- a/engine/src/linux.stubs
+++ b/engine/src/linux.stubs
@@ -72,6 +72,9 @@ pangoft2 libpangoft2-1.0.so.0
 
 pangocairo libpangocairo-1.0.so.0
 	pango_cairo_font_map_get_default: () -> (pointer)
+	pango_cairo_create_layout: (pointer) -> (pointer)
+	pango_cairo_update_layout: (pointer, pointer) -> ()
+	pango_cairo_show_layout: (pointer, pointer) -> ()
 
 pango libpango-1.0.so.0
 	pango_context_new: () -> (pointer)
@@ -91,6 +94,7 @@ pango libpango-1.0.so.0
 	pango_layout_set_font_description: (pointer, pointer) -> ()
 	pango_layout_get_extents: (pointer, pointer, pointer) -> ()
 	pango_layout_get_pixel_extents: (pointer, pointer, pointer) -> ()
+	pango_layout_get_pixel_size: (pointer, pointer, pointer) -> ()
 	pango_layout_get_iter: (pointer) -> (pointer)
 	pango_layout_iter_free: (pointer) -> ()
 	pango_layout_iter_get_run: (pointer) -> (pointer)
@@ -301,6 +305,7 @@ gdk libgdk-x11-2.0.so.0
 	gdk_keyval_to_unicode: (integer) -> (integer)
 	gdk_rgb_find_color: (pointer, pointer) -> ()
 	gdk_drawable_get_colormap: (pointer) -> (pointer)
+	gdk_drawable_get_visual: (pointer) -> (pointer)
 	gdk_window_set_background: (pointer, pointer, integer) -> ()
 	gdk_notify_startup_complete: () -> ()
 	gdk_window_ensure_native: (pointer) -> (integer)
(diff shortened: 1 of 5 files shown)


emily-elizabeth:
image

emily-elizabeth:
Still crashing when closing the stack window, but it might not even be our code causing the crash.

emily-elizabeth:
Looks like the crash has been fixed, can I get someone to double-check

mwieder:
Looks good to me as far as it goes. No segfault crashing, toolbar seems to behave. I assume the icons thing has to wait until after that PR is established because right now it does nothing. Otherwise good to go.

emily-elizabeth:
You don't get a gear icon for the first toolbar item?

mwieder:
I see a gear icon when loading the stack. No other changes. The 'tool icon' button doesn't add icons to  the New or Open menuItems.

emily-elizabeth:
So long as you see that gear icon, that means the icons are working. The only other "icon" is those weird macOS-only symbols for the Save toolbar item

mwieder:
Ah, OK. So setting the toolIcon fails silently otherwise, as in the code behind the 'tool icon' button.
No problems then.

emily-elizabeth:
So is this good to go?

engine/src/linux.stubs
```diff
@@ -353,11 +360,29 @@ gdk_pixbuf libgdk_pixbuf-2.0.so.0
...
 	cairo_clip: (pointer) -> ()
 	cairo_paint: (pointer) -> ()
-	cairo_rectangle: (pointer, integer, integer, integer, integer) -> ()
+	cairo_rectangle: (pointer, double, double, double, double) -> ()
 	cairo_fill: (pointer) -> ()
+	cairo_pattern_create_linear: (double, double, double, double) -> (pointer)
+	cairo_pattern_add_color_stop_rgb: (pointer, double, double, double, double) -> ()
+	cairo_set_source: (pointer, pointer) -> ()
+	cairo_pattern_destroy: (pointer) -> ()
+	cairo_save: (pointer) -> ()
```
mwieder, line 372: Some seriously weird code here, no?

mwieder (approved):
Looks good.

engine/src/linux.stubs
emily-elizabeth: So long as it work 😄

PR #300, opened 2026-05-31
Fix/lin popover

emily-elizabeth:
popover windows now have a shadow and disappear when switching applications. there are window decorations, but we will live with them until the GTK3 migration is complete and we can use GTKPopover instead.


mwieder:
Looks good here.

mwieder (approved):
Looks good here. I'm getting a proper closing message clicking on the desktop as well as closing the parent stack.

PR #304, opened 2026-05-31
Fix/lin dark mode

emily-elizabeth:
fix all the outstanding dark mode issues on Linux
- SystemAppearanceChanged message gets fired
- SystemAppearanceChanges parameters are correct
- SystemWindowColor/SystemTextColor properties return correct colours
- IDE updates with system appearance changes

### DIFF
--- a/engine/src/desktop.cpp
+++ b/engine/src/desktop.cpp
@@ -49,8 +49,13 @@ extern "C" __attribute__((weak)) void MCplatformGetLabelColor(char *, size_t) {}
 extern "C" bool MCplatformIsDarkMode(void);
 extern "C" void MCplatformGetWindowBackgroundColor(char *p_buf, size_t p_buflen);
 extern "C" void MCplatformGetLabelColor(char *p_buf, size_t p_buflen);
+#elif defined(TARGET_PLATFORM_LINUX)
+// Real implementations provided by lnxgtktheme.cpp (GTK-based).
+extern "C" bool MCplatformIsDarkMode(void);
+extern "C" void MCplatformGetWindowBackgroundColor(char *p_buf, size_t p_buflen);
+extern "C" void MCplatformGetLabelColor(char *p_buf, size_t p_buflen);
 #else
-// Stub fallbacks for Linux and other non-Mac, non-Windows platforms.
+// Stub fallbacks for all other platforms.
 extern "C" bool MCplatformIsDarkMode(void) { return false; }
 extern "C" void MCplatformGetWindowBackgroundColor(char *, size_t) {}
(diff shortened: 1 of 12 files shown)


emily-elizabeth:
The only issue remaining is the script editor not switching with the system appearance until it is closed and reopened

emily-elizabeth:
Can you make sure I didn't miss anything (except that outstanding script editor issue)

emily-elizabeth:
merge this PR before 314

mwieder (commented):
Um. Lots of problems in linux.
Multiple (too many to count) instances of "g_object_get_is_valid_property: object class 'GtkSettings' has no property named 'gtk-application-prefer-dark-theme'"
Changing to/from dark mode while the IDE is active crashes with signal 11.

emily-elizabeth:
I see the messages, but I'm not getting a crash when switching the system appearance unless I select an image and open the property inspector.

mwieder:
OK - I can narrow the segfault down to third-party plugins. That's still not good and I'll see if I can narrow this down a bit. From gdb I can see the error is caused at g_type_check_instance, but that's deep in the system and not much help at this point. FWIW I see the segfault on quitting even without changing dark/light mode, so this may be a separate issue except for the IDE quitting on changing modes.

mwieder:
Yep - it's my PowerTools plugin. I'll need to take a look at what here is causing the segfault. I do need to say that there's no problem switching modes in LC 10.0.5, only in HXT. Otherwise I'd just say this is good to go.

mwieder:
Well, that was quick. Any stack with a backgroundcolor set will trigger the segfault.
Recipe:
create a new stack
toggle the mode to ensure it doesn't crash
set the backgroundcolor of the stack to anything
toggle the mode again and get the segfault

emily-elizabeth:
Seeing this on Linux
macOS is free of the issue
will check Windows tomorrow

emily-elizabeth:
Windows is clear of this issue

mwieder:
What led me to this was the call to MCplatformGetWindowBackgroundColor. Figured I should try it.

emily-elizabeth:
Looks like this is good for testing, when you have a moment.

mwieder:
No problems switching modes. Still a segfault at g_type_check_instance on exit if any stack has a backgroundcolor.

emily-elizabeth:
Sorry, I missed that one. Working on it now.

emily-elizabeth:
Okay, third times a charm? 😉

mwieder:
Yep. Looks good here.

emily-elizabeth:
It fixed another bug I was tracking down, so I'm happy about that

PR #307, opened 2026-06-01
Fix/answer dialog return key

### LINKED ISSUE #305: default button in answer dialogs do not trigger when pressing enter or return key
> default button in answer dialogs do not trigger when pressing enter or return key

emily-elizabeth:
Restored the `revAnswerDialog.rev` from 0.9.13 and the functionality returned


emily-elizabeth:
tested on macOS and it works, will test Windows and Linux tomorrow

emily-elizabeth:
Test on macOS, Windows and Linux and it seems to be working

BerndN:
I tested the binary stack "Answer Dialog" using HXT 0.9.14
It works flawlessly.

However it does not change appearance when using dark mode.

Adding to the preOpenStack handler right on top after the declaration of variables

   ------------------------
   // HXT dark mode
   if the systemAppearance is "dark" then
      set the backColor of me to "50,50,50"
      set the foreColor of me to "white"
      repeat with i = 1 to the number of buttons of me
         set the foreColor of button i of me to "black"
      end repeat
   else
      set the backColor of me to "235,235,235"
      set the foreColor of me to "black"
   end if
   ------------------------

makes it dark mode aware and works in my testing.

emily-elizabeth:
I'm not seeing any issues with dark mode.
- launch HyperXTalk in light mode
- answer dialog from the message box
- change the system appearance to dark mode

- launch HyperXTalk in dark mode
- answer dialog from the message box

image

emily-elizabeth:
You shouldn't use "black" or other hardcoded colours, but use the systemWindowColor and systemTextColor properties to get the system colours

BerndN:
> You shouldn't use "black" or other hardcoded colours, but use the systemWindowColor and systemTextColor properties to get the system colours.

In my experience they always are:
#000000
#ececec
regardless of dark/light mode setting

I would like to avoid closing and reopening HXT but rather change the dark/light mode when the system changes, just like other apps do it. Maybe I am missing something.

emily-elizabeth:
The windows and text should automatically switch when the system appearance changes, no need for the `SystemAppearanceChanged` handler. If you're not seeing that, then there's something going on that needs to be fixed. Also the systemWindowColor and systemTextColor get the current system colours on all three platforms so that way HXT looks right

BerndN:
Ok, then I would say it is okay.

PR #316, opened 2026-06-01
Updated gitignore

mwieder:


mwieder:
Simple to resolve - I think that leading slash should not be there.

emily-elizabeth (approved):

bwmilby:
`**/extensions/*/*/api.lcdoc`
That needs to be included.  We should not be including derived files in the repo.  The `api.lcdoc` files used to build an app are generated during compile and not the ones that live next to the `.lcb` source anyway.  It is only when launching the app from the repo that those files get used.
(there is another PR that does add it, not sure if it will end up being a conflict)

We really should take a careful look at the ignore file from the original repo.  I think we should include reasoning whenever we remove something that was originally present.  A good example would be the binaries needed for building apps for other architectures.

mwieder:
Agreed. There's a long pedigree for this .gitignore, and I think it needs some TLC.

emily-elizabeth:
If the file is being tracked adding it to gitignore still doesn't work

mwieder:
doesn't git remove stop it from tracking?

emily-elizabeth:
I have no idea how to stop tracking it. The last time I followed instructions from the internet, it took me four days to get the Windows package building again

bwmilby:
`git rm --cached filename` will remove a file from the repo and leave it local but the problem with this particular file is that a specific addition is required for windows builds.  I don't think we can remove it from being tracked without doing something else to ensure win builds don't break.

The good thing is that if you review the PR in GitHub, there will be a nice big red section that tells you not to commit the file.

It's pretty easy to tell from the file that nothing changed needing an update (linux just removed the win addition).

PR #322, opened 2026-06-03
user plugins and schemes folders are created on launch

### LINKED ISSUE #308: Plugins Folder Not Created
Upon initial launch, the Plugins folder is not created.  Empty "Libraries" and "Extensions" folders exist.

emily-elizabeth:
~/Documents/HyperXTalk/Plugins
~/Documents/HyperXTalk/Schemes

folders are created when HyperXTalk launches

### DIFF
--- a/ide/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/ide/Toolset/libraries/revidelibrary.8.livecodescript
@@ -4056,7 +4056,6 @@ local tIDEFolder
 put revIDESpecialFolderPath("IDE") into tIDEFolder
 */
 function revIDESpecialFolderPath pKey, pParam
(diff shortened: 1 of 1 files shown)


bwmilby:
Instead of defining a local variable, you could just use `get` for those 2 calls

```
   // HXT - ensure folders are created
   get revIDESpecialFolderPath("user plugins")
   get revIDESpecialFolderPath("user schemes")
```

BerndN:
It works ok, except for the frequent debugging output to the message box

function revIDESpecialFolderPath pKey, pParam
   put pKey, pParam & LF after msg  // HXT <----

emily-elizabeth:
Corrected both of those issues

PR #323, opened 2026-06-03
[[Issue 302]] Update macScrollEnhancement support for Project Browser

### LINKED ISSUE #302: [mac] Project Browser does not scroll when cursor is over text
> In HXT 0.9.14 the project browser does not scroll anymore. Just manually dragging the scrollbar with the mouse moves the PB.

### LINKED ISSUE #325: Scroll events outside of a target field
Given a scrollabel field that has no focus and the mouse is outside of the field to the left or at the top and the mouse is scrolled scrolls the fields content.
The scrollable area to the left of the field is about 100 pixel wide and includes about 80-90 percent of the height to the bottom.
I find it confusing if one accidentally scrolls outside of a non-focused scrollable field and the field scrolls.

This was not the case for HXT 10 and is not present in LC.

If the field is focused then this also happens but that has always been the case.

Furthermore when the scrolling enhancement is active and the field is focused and the mousescroll is to the left of the field adaptive scrolling is only working in the upper part whereas "raw", rapid scrolling is triggered in the lower part left of the foused field.
This is particularly bizarre when the mouse is scrolling over the focused field, the field is scrolling and one moves the mouse out of the scrolling field to the left in the lower part of the field: "raw" scrolling happens and one did not expect this.

bwmilby:
Special case data views like the Project Browser to use `viewprop["vscroll"]` to set the `vscroll`.  Bounds checking is performed by the data view so just pass raw updated scroll value.  Only updates if deltaY > .5 to allow tooltips to show.

Handle the case where the mouse is over a card and there is a selected field.

Update code to use `the long id of the mousecontrol` which eliminates the need to use `do script`.

Removed no longer needed internal helpers.

Removed commented out code.

Replaced `‚Äî` with `-` in the script.

Changed version to 1.0.1 due to the changes in how it works.

Closes #302 
Closes #325

### DIFF
--- a/extensions/script-libraries/macscrollenhancement2/macscrollenhancement2.livecodescript
+++ b/extensions/script-libraries/macscrollenhancement2/macscrollenhancement2.livecodescript
@@ -26,19 +26,19 @@ Author: HyperXTalk
 
 Type: library
 
-Version: 1.0.0
+Version: 1.0.1
 
 Description:
   Generic smooth-scroll library for HyperXTalk.
 
   Intercepts precise trackpad scroll events application-wide via a C
   NSEvent monitor and applies accumulated pixel deltas to whichever
-  scrollable LiveCode control is currently under the mouse ‚Äî fields,
+  scrollable HyperXTalk control is currently under the mouse - fields,
   groups with scrollbars, etc.
 
   Call seScrollEnhancementEnable from any openStack handler that should
   benefit from smooth scrolling. Call seScrollEnhancementDisable from the
-  matching closeStack. Reference-counted ‚Äî the poll keeps running until
+  matching closeStack. Reference-counted - the poll keeps running until
   every caller has disabled it.
 */
 
@@ -72,7 +72,7 @@ end seScrollEnhancementEnable
 /**
 seScrollEnhancementDisable
 Stop smooth scrolling. Call from closeStack (or equivalent).
-Reference-counted ‚Äî poll stops only when every caller has disabled.
+Reference-counted - poll stops only when every caller has disabled.
 */
 command seScrollEnhancementDisable
    if the platform is not "MacOS" then exit seScrollEnhancementDisable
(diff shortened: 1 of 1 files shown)


BerndN:
I tested the patch (copied it from here and pasted it into "macscrollenhancement2" in the IDE. The patch scrolls the project browser very well. However tooltips do not show up anymore. Clicking in the footer to do a "refresh" stops scrolling, likewise opening a script by clicking on it in PB stops scrolling. Clicking into the header of the PB stops scrolling. If there is a selection marked in PB to indicate a selected object and the mouse hovers over it the scrolling stops. If scrolling stops then nothing scrolls at all anywhere. (This is the same as in the original enhancement)


To help with testing I made a stack with 1 button, the script of the button to restart scroll enhancement since I saw no other way except for a restart scrolling.
-----------------
on mouseUp
   local tStack
   put "com.hyperxtalk.library.macscrollenhancement2" into tStack
   send "seScrollEnhancementDisable" to stack tStack
   send "seScrollEnhancementEnable" to stack tStack
end mouseUp
-----------------

bwmilby:
Actually you can just manually trigger `seScrollPoll` to restart scrolling (don't have to disable/reenable).

I do see that tooltips are not showing in the Project Browser but I'm not getting any of the other halts that you are reporting.  The reason they were showing before is that this script was only picking up the fields.  The other controls when walked up didn't trigger the script.

I just updated my commit with more generalized data view handling.

When you say it stops working, is that accompanied by an error show in the script?  Usually whenever scrolling stopped working it was after an error that interrupted the `seScrollPoll` script.

BerndN:
> Actually you can just manually trigger seScrollPoll

That works very well from the message box. Thank you very much.

bwmilby:
I just updated the PR commit.  I added a throttle (like the original code) to the data view handling (deltaY > .5).  In my testing, as long as the Project Browser window is in front of the script editor then tooltips show.  If behind, then they do not.  I believe this is expected behavior.

BerndN:
I tried this:

```
if the dvcontrol of tScrollTarget is not empty then
         // bounds checking performed by data view so just use raw delta
         
         // let tooltips popUp if not scrolling
         if tDeltaY is 0 and tDeltaX is 0 then
            send "seScrollPoll" to me in 16 milliseconds
            exit seScrollPoll
         end if
         set the viewprop["vscroll"] of tScrollTarget to \
```

and it worked on Project Browser when PB is in front.

bwmilby:
That is essentially what I ended up doing.  I only check Y since X isn't implemented for the data view.  I mirrored the code further down.  I also just updated the push to handle selected fields.
```
         if abs(tDeltaY) > 0.5 then
            set the viewprop["vscroll"] of tScrollTarget to \
                  (the viewprop["vscroll"] of tScrollTarget) + tDeltaY
         end if
         send "seScrollPoll" to me in 16 milliseconds
         exit seScrollPoll
```

BerndN:
That works well now, still a scrollable area outside of a non-focused field, now at the left side near the top of the field and on the top of the non-focused field a small area to the left edge.

bwmilby:
How large of an area are you seeing?  I can position the cursor one pixel to the top or left and get it to scroll but any further and it does nothing.  This particular test was with the previous release and a copy/paste of the updated script.  I've also been doing most other tests with a fresh compile on this branch directly off of main.

emily-elizabeth (approved):
Verified the project browser scrolls properly when the mouse cursor is over a text field

BerndN:
Make the field 500 pixel height and 400 pixel width, then you see it.
At 160 height and 260 width I do not see it.

PR #327, opened 2026-06-04
[[Issue 325]] macscrollenhancement mouselwc fix and widget support

### LINKED ISSUE #325: Scroll events outside of a target field

### LINKED ISSUE #326: [feature] Add widget support to macscrollenhancement
Currently widgets are not handled in macscrollenhancement.

I have some code that can enable support for any widget with hscrollbar and/or vscrollbar properties.  The browser widget must be excluded since it is a native control and the OS handles scrolling.  This would be the case for any other native controls as well (like a Mac text field).

The tree view widget is the one where I can see this being useful.  For short lists it will not be a big deal but if there is a good bit of data then the enhancement will be a good benefit.

The question is should it be all widgets with an exclusion list or should it only be specific widgets included?  I think using the scrollbar properties will make the exclusion list the better option, but wanted additional input.

If you want to take a look at what I'm working on with it, my branch is here:
https://github.com/bwmilby/HyperXTalk/tree/bwm-macscrollenhancement_for_widgets

bwmilby:
When the cursor was outside of any control (`mousecontrol` empty), the code was walking up the controls in the stack looking for a hit.  The `mouseloc` was returning global coordinates which caused it to match controls that it should not.  Added a function to convert the `mouseloc` to be relative to the `mousestack`.

Add support for scrolling widgets

Exclude browser widget (native control, OS handles)

Closes #325 
Closes #326

### DIFF
--- a/extensions/script-libraries/macscrollenhancement2/macscrollenhancement2.livecodescript
+++ b/extensions/script-libraries/macscrollenhancement2/macscrollenhancement2.livecodescript
@@ -118,7 +118,7 @@ function seGetScrollTarget
       local tN
       put the number of controls of the mouseStack into tN
       local tMouseLoc
-      put the mouseLoc into tMouseLoc
+      put _seGetRelativeMouseLoc() into tMouseLoc
       repeat with i = tN down to 1
          local tCandidate, tRect
          put the long id of control i of the mouseStack into tCandidate
@@ -139,13 +139,18 @@ function seGetScrollTarget
          -- Skip fields inside the line-number gutter.
          local tParent
          put the long id of the owner of tControl into tParent
-         if word 1 of tParent is "group" and the short name of tParent is "Gutter" then
+         if the name of tParent is ("group" && quote & "Gutter" & quote) then
             put tParent into tControl   -- continue walking up from the gutter group
          else
             return tControl
          end if
          
-      else if tKind is "group" then
+      else if tKind is "group" or tKind is "widget" then
+         -- Exclude the browser widget since it is a native control handled by the OS
+         if tKind is "widget" and the kind of tControl ends with "browser" then
+            return empty
+         end if
+
          local tHasScrollbar
          try
             put the vScrollbar of tControl into tHasScrollbar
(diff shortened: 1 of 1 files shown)


emily-elizabeth:
Still seeing the issue the first time in run mode, but after toggle modes it goes away

emily-elizabeth:
Issue also happens in a standalone, until you switch applications and switch back

bwmilby:
[scroll_test.hyperxtalk.zip](https://github.com/user-attachments/files/28622408/scroll_test.hyperxtalk.zip)
I did a `make config-mac` then `make compile-mac` then `hxide` (launches from `Mac-bin`) then opened attached stack.  Field selected so it scrolled when over card.  Switched to run mode.  Clicked the card to deselect field and it no longer scrolled when outside.  Built it as an app and same thing happened - did not notice scrolling that shouldn't happen.

Using the same stack opened in previous release I can get it to scroll both fields when none selected at spots near the top and left of the window.

Can I get your recipe to see if I can reproduce from my branch?

emily-elizabeth:
- rebuilt HXT based on this PR
- create a new stack, add a field, enlarge the field
- switch to run mode
- add text to the field so that it scrolls
- move mouse outside of text field (to the top-left of the field)
- scroll the trackpad and watch the text field scroll
- toggle the mode (browse, run) and the issue goes away

bwmilby:
That sounds like the programmed behavior... if the field is selected then scroll works anywhere on the card where there is no other control.  I tried in a Create DP and it works the same.  The one difference is that if the text field has no scrollbar, then it will not scroll at all in Create DB but it does the way we have this coded.  You can still move the text by arrowing up/down inside the field regardless of presence of scrollbars.

PR #335, opened 2026-06-05
Updated libgif to version 6.1.3

mwieder:
Pushed the version to 6.1.3. Minor update.


emily-elizabeth:
We'll have to upgrade the other two platforms too and seeing as we just updated to 6.1.2 this will have to wait for a bit

mwieder:
OK - no rush. Just validating that it compiles and runs properly here.

PR #337, opened 2026-06-05
urlencode space as %20

mwieder:
Per the conversation in the forum, this converts spaces to '%20' instead of '+'. The urldecode function is unchanged and works properly for either conversion. This brings url encoding up to date with RFC3986. This is a simple change to the lookup table and I don't imagine any downside to it. Indeed there are uses for '+' in urls separate from an encoded space. Gmail addresses, for example, can have embedded plus signs as a key for email filters.


emily-elizabeth:
tested on macOS and Windows, just waiting for Linux to finish compiling

emily-elizabeth:
tested on Linux

emily-elizabeth (approved):

PR #338, opened 2026-06-06
fix(scripting): add JavaScript alternate language support on arm64 macOS

### LINKED ISSUE #324: do ... as "JavaScript" — “Language not installed”
> Running HXT 0.9.14 on macOS (Apple Silicon). Calling do "encodeURI(...)" as "JavaScript" throws:
> 
> > alternate language not found
>
> My guess is that on macOS this routes through OSA / JXA (JavaScript for Automation), which is a completely different engine from a browser JS context and doesn’t expose browser globals like encodeURI. So either JXA support was never reliable for this use case, or something changed in the OSA bridge.
> 
> Questions:
> 
> - Is do ... as "JavaScript" on macOS officially supported?
> - Is there a recommended in-engine alternative for URL encoding a string?
>
> For now I’ve routed around it with a custom bbencode function doing percent-encoding manually, but it would be useful to know the intended behavior.

emily-elizabeth:
do "..." as "JavaScript" was throwing alternate language not found on Apple Silicon because the arm64 rewrite of getosacomponents() hardcoded only "applescript" — Component Manager (FindNextComponent) being unavailable on arm64 meant JavaScript was never registered as an available language. 

Fix adds a dedicated JavaScript execution path in DoAlternateLanguage in dskmac.mm that runs the script through JSContext (JavaScriptCore's modern Obj-C API, arm64-native). This gives a full JS environment including standard built-ins like encodeURI and encodeURIComponent. getosacomponents() is updated to register "javascript" as a second supported language so the alternateLanguages reports it correctly.

### DIFF
--- a/engine/src/dskmac.cpp
+++ b/engine/src/dskmac.cpp
@@ -6,6 +6,7 @@
 #include <CoreServices/CoreServices.h>
 #include <CoreFoundation/CoreFoundation.h>
 #import <AppKit/AppKit.h>    // NSRunningApplication, NSWorkspace, NSAppleScript
+#import <JavaScriptCore/JavaScriptCore.h>  // JSContext for do...as "JavaScript"
 
 #include "parsedef.h"
 #include "filedefs.h"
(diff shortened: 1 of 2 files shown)


emily-elizabeth:
works on macOS and no conflicts on Windows or Linux

PR #339, opened 2026-06-06
perf(json): fix O(n²) token accumulation in JsonImport

### LINKED ISSUE #296: O(n²) JsonImport: per-char string-token accumulation in json.lcb
**Summary:** `JsonImport` parses in O(n²) for large string/number tokens. `extensions/libraries/json/json.lcb` tokenises with `repeat for each char tChar in pJson` (line 203) and accumulates token text via per-char `put tChar after tTokenValue` (lines 280, 284, 321, and the escape/number branches). Because LCB strings are immutable / copy-on-write, each append copies the whole accumulated token, so a token of M chars costs Σ ∝ O(M²).

```
-- extensions/libraries/json/json.lcb:203
repeat for each char tChar in pJson
   ...
   -- :280 (also :284, :321, and the \n \r \t \u branches)
   put tChar after tTokenValue
```

**Repro:** `JsonImport` of a document containing one large string value, e.g. `{"data":"<1 MB of text>"}`. Parse time grows quadratically with the value length.

**Impact:** Any use of the bundled JSON library to import documents with large string/number values (config blobs, base64 data, RPC bodies) is slow.

**Suggested fix:** Accumulate token characters into a buffer/list and join once at the token boundary instead of repeated `put ... after` on an immutable string. That makes tokenisation O(n).

Related: `MCSocket::readsome` in `engine/src/opensslsocket.cpp` has the same quadratic shape for socket reads (filed separately).

emily-elizabeth:
JsonImport was building string and number token values by appending one character at a time with put tChar after tTokenValue. Because LCB strings are immutable/copy-on-write, each append copies the entire accumulated token — making tokenisation O(M²) for a token of M characters. A document containing a single 1 MB string value would parse quadratically.

Fix replaces per-character string concatenation with a List accumulator (tTokenChars). Characters are pushed onto the list during scanning and combined into a string once at the token boundary via combine tTokenChars with "" into tTokenValue. Literals (true, false, null) combine at their length check; strings and numbers combine just before the parser stage. The outer repeat for each char loop remains O(n) as before.

File changed: extensions/libraries/json/json.lcb


emily-elizabeth:
[jsonImporter.hyperxtalk.zip](https://github.com/user-attachments/files/28669452/jsonImporter.hyperxtalk.zip)
[users-10.json](https://github.com/user-attachments/files/28669332/users-10.json)
[logs-10000.json](https://github.com/user-attachments/files/28669333/logs-10000.json)

emily-elizabeth:
tested on macOS and Linux and getting around five seconds for the 10,000 log file (was 36 seconds previously)

emily-elizabeth:
tested on Windows - two seconds

emily-elizabeth:
The JSON extension now uses yyjson to help speed up JSON imports

bwmilby:
As a curiosity, I tried importing the log file using the existing LCB and PhotonJSON on my M4 Air.  Times came to 11.4 seconds for LCB and 0.6 seconds for PhotonJSON (LCS only).  Compiled from the new branch it comes in as 1.2 seconds for LCB/yyjson.  At this point I would say this is no longer a "livecode" extension - I'd change the library name to hyperxtalk origin.

emily-elizabeth:
I wonder why PhotoJSON is so fast?

emily-elizabeth:
Good idea to change the origin to hyperxtalk

emily-elizabeth:
okay, tested on macOS, Windows and Linux. Only the com.hyperxtalk.library.json folder in the extensions folder

PR #340, opened 2026-06-07
Feature/global hotkeys clean

emily-elizabeth:
allow your application to respond to global hotkeys


emily-elizabeth:
tested on macOS

engine/src/executionerrors.h
bwmilby, line 1: I think renumbering errors is probably not the best practice.

emily-elizabeth: I'll keep that in mind for the future

bwmilby:
I would suggest that after any fixes are done that all of the commits be squashed prior to being merged.
We should change the header to all of the files - we should no be using the LiveCode boilerplate for new additions.

engine/src/executionerrors.h
bwmilby: (same thing for parse issues)

emily-elizabeth:
> I would suggest that after any fixes are done that all of the commits be squashed prior to being merged. 

How do we do that? Some of us are newbs at GitHub ;-)

> We should change the header to all of the files - we should no be using the LiveCode boilerplate for new additions.

What do you mean by that?

emily-elizabeth:
tested on Windows

bwmilby:
> > I would suggest that after any fixes are done that all of the commits be squashed prior to being merged. 
> 
> 
> 
> How do we do that? Some of us are newbs at GitHub ;-)
> 
> 
Easiest way to do it manually is using the GitHub Desktop app on a Mac (that is what I use).  You can just select all of the other commits and drag them to the first one to squash.  You can edit the commit message to include all changes and remove any back-out comments since you are not including the in and out in the history.

After making the changes, you will need to force-push the commit back to GitHub.
> > We should change the header to all of the files - we should no be using the LiveCode boilerplate for new additions.
> 
> 
> 
> What do you mean by that?
> 
> 

The 5 new code files all start with the LiveCode copyright line with years in the past.

bwmilby:
Makefile target should be 'sync-toolset-mac' since it is OS specific.

emily-elizabeth:
> The 5 new code files all start with the LiveCode copyright line with years in the past.

oh, that is because they are really old and were created before I removed the copyrights. I also need to change the Introduced flags in the dictionary

bwmilby:
> Makefile target should be 'sync-toolset-mac' since it is OS specific.

Actually, I don't think that target needs to be there at all.  If you launch HyperXTalk from `mac-bin` it does not pull from the app bundle at all.  That only happens once the package is moved outside of the repo location.  Easy way to verify is to put the following in the message box when launching right after a compile:
```
put the long name of stack revdocsparser
```
In other words, you are not really testing the packaged app while it is in the original repo location.  If you want to test it fully, it needs to be moved outside of the repo.  Once done, then you can't use make to update the files inside.  Relaunching is enough to pull in any changes to `livecodescript` files.  Recompile will be needed for LCB changes.

emily-elizabeth:
For testing `systemd-run --user --scope --unit="app-hyperxtalk-$(date +%s).scope" ./HyperXTalk`

emily-elizabeth:
Brian, can you test this on Linux to see if works for you. If it doesn't, can you post what you're using? Tested on Ubuntu 25.04 under Wayland

bwmilby:
I'm starting a compile now.  I'm running Ubuntu 24.04.04 on an older HP i5 laptop.

bwmilby:
It works some.  I'm not seeing errors, but just non-responsiveness depending on what is active.  Terminal, Files, Firefox, desktop ignore it.  So far only Code (Microsoft) recognizes it (besides HyperXTalk-have 0.9.14 in front and it brings the compiled one forward).

emily-elizabeth:
Okay, thank you for testing. I guess it's "ready" even though Linux applications seem to have issues with it (not too much I can about that)

bwmilby:
Just be sure to do a "squash" when merging because the history doesn't need to show things like debug adds and removals and things that were tried and then revised.

emily-elizabeth:
There I think I did that correctly

PR #344, opened 2026-06-12
better localized date/time format

### LINKED ISSUE #329: Incorrect values returned by system date and system time functions
The system date and system time functions return incorrect values.

This is not a HyperXTalk-specific issue. The same behavior can be reproduced in the latest Community Edition of LiveCode.
The attached screenshots from LiveCode 10 demonstrate the expected behavior, where the issue has been corrected and the functions return the correct values.

For testing, I’ve included Sarah’s "Dates and Time" stack as well.

Image

Image

[Dates and Time.hyperxtalk.zip](https://github.com/user-attachments/files/28626243/Dates.and.Time.hyperxtalk.zip)

mwieder:
Possible fix for https://github.com/emily-elizabeth/HyperXTalk/issues/329. It's working for me in all situations with my limited ability to test with various different localizations, but obviously could use more validation on different platforms and locales.
Note that the logic in the file is a bit screwy and hard to follow but is necessary to maintain compatibility with different OS-dependent platform localizations.

### DIFF
--- a/engine/src/date.cpp
+++ b/engine/src/date.cpp
@@ -776,25 +776,28 @@ void MCD_time(MCExecContext &ctxt, Properties p_format, MCStringRef &r_time)
 	switch(t_length)
 	{
 		case P_LONG:	// long time is always 24-hour format
-		case P_INTERNET:
 			break;
 		default:
-			t_length = P_SHORT;
+			t_length = P_ABBREVIATE;
 	}
 
 	const MCDateTimeLocale *t_locale;
-	if (t_use_system && P_INTERNET != t_length)
+	if (t_use_system)
 		t_locale = MCS_getdatetimelocale();
 	else
 		t_locale = g_basic_locale;
 
 	MCStringRef t_date_format;
-	if (P_INTERNET == t_length)
-		t_date_format = MCSTR(s_internet_date_format);
-	else if (MCtwelvetime)
-		t_date_format = t_locale -> time_formats[t_length - P_SHORT];
+	if (MCtwelvetime)
+	{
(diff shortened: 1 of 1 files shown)


bwmilby:
Code looks reasonable.  A few changes aren't strictly necessary (if -> switch for example).  Compiles fine for me.

mwieder:
Right, but the if->switch mods made the code easier for me to understand what's going on. The bit about subtracting enums is still murky to me, and then there's the 'oddity' comment from LC.

mwieder:
So what do we do with this? I realize not everyone can test it at the DIY level. Do we merge it and send it out for validation and then revert it if it's not universally correct? Wouldn't be my favorite way to move forward, but I'm not sure what else to suggest. Make a special test build for Rolf?

bwmilby:
I'd say if we can test on our seats and it looks fine, then put it in .15 and announce it as a change to test.  We don't have an extensive group that can pull and test from the repo so I think this is a good way.  Just be sure that we have a single commit we can reverse if needed (if any conflicts arise before merge, rebase and squash instead of fix via merge).

Having said that, I just tested on my Mac and it does not seem to make any difference.  I still see English.

mwieder:
/sigh/
I seem to have pushed *before* committing the latest changes. Pushed again. Hopefully the right stuff this time.
The problem stems mostly from the fact that not all locales have twelve-hour time and so return a system date format that can't be parsed. I put a kludge into the time function to see if a reasonable format string is returned for a twelve-hour system time and default to 24-hour time if not. I see the LC team struggled a bit with this as well.

`// We have to make time parsing more permissive. This is because of the introduction of system
// times when nobody's scripts are ready for it.
`

PR #346, opened 2026-06-14
Fix/vlc audio only playback clean

emily-elizabeth:
Audio only players would not play due to a height check, so the height check is only used when the media has a video track. Fixed for macOS and Windows. Linux was not affected.

Had to make a new PR due to the changes made and it not compiling cleanly on Windows


emily-elizabeth:
tested on macOS

emily-elizabeth:
tested on Linux

emily-elizabeth:
tested on Windows

PR #349, opened 2026-06-15
[[Issue 341]] Add `the properties` support for widgets

### LINKED ISSUE #341: the properties returns empty for widgets
`the properties` returns an array of the properties for objects.
For any widget, it returns an empty array.  You are able to set the properties of widgets.
I believe that for the essentials (rect, visible, etc.), they should be included in what it returned for widgets.  I would also think that including what is provided via "export widget x to array y" should be included as well, but would settle for the base properties.

This was reported as an issue to LC and deemed potentially not a bug (still sitting in "Expert Review":
https://quality.livecode.com/show_bug.cgi?id=23670

Having skimmed through the discussion on the use-list at the time, I still think it would be useful to have the basics exported.

Why am I asking now?  I'd like to be able to export stack data to a JSON file so that differences can be tracked via GitHub (or some other VCS).  We have a few tools available that attempt to do that sort of thing, but using `the properties` should make it easier and the code cleaner.  One of my goals would be that the export would be something that could be understood by a human.  One tool, SmartCrumbs, uses GUIDs for everything and you cannot just look at a script file and figure our where it came from (no metadata present in the file - you would need to look up the GUID in the JSON file and track down the object name).

bwmilby:
Add basic support to get `the properties` for widgets.  Set has always worked.

```
static const PropList widgetprops[] =
    {
        {"altId", P_ALT_ID},
        {"behavior", P_PARENT_SCRIPT},
        {"blendLevel", P_BLEND_LEVEL},
        {"disabled", P_DISABLED},
        {"id", P_ID},
        {"ink", P_INK},
        {"label", P_LABEL},
        {"layer", P_LAYER},
        {"layerMode", P_LAYER_MODE},
        {"lockLoc", P_LOCK_LOCATION},
        {"name", P_SHORT_NAME},
        {"rect", P_RECTANGLE},
        {"visible", P_VISIBLE},
    };
```

Still will need to `export widget x to array y` to get the state data.

Closes #341

### DIFF
--- a/engine/src/exec-interface-object.cpp
+++ b/engine/src/exec-interface-object.cpp
@@ -757,6 +757,23 @@ static const PropList videoclipprops[] =
         {"scale", P_SCALE},
     };
 
+static const PropList widgetprops[] =
+    {
+        {"altId", P_ALT_ID},
+        {"behavior", P_PARENT_SCRIPT},
+        {"blendLevel", P_BLEND_LEVEL},
+        {"disabled", P_DISABLED},
+        {"id", P_ID},
+        {"ink", P_INK},
+        {"label", P_LABEL},
+        {"layer", P_LAYER},
+        {"layerMode", P_LAYER_MODE},
+        {"lockLoc", P_LOCK_LOCATION},
+        {"name", P_SHORT_NAME},
+        {"rect", P_RECTANGLE},
+        {"visible", P_VISIBLE},
+    };
+
 ////////////////////////////////////////////////////////////////////////////////
 
 void MCInterfaceLayerParse(MCExecContext& ctxt, MCStringRef p_input, MCInterfaceLayer& r_output)
@@ -3382,9 +3399,8 @@ void MCObject::DoGetProperties(MCExecContext& ctxt, uint32_t part, bool p_effect
         tablesize = ELEMENTS(toolbarprops);
         break;
     case CT_WIDGET:
-		table = NULL;
-		tablesize = 0;
-        // WIDGET-TODO: Implement properties
+        table = widgetprops;
+        tablesize = ELEMENTS(widgetprops);
         break;
 	default:
 		return;
--- a/engine/src/widget.cpp
+++ b/engine/src/widget.cpp
@@ -157,11 +157,6 @@ const char *MCWidget::gettypestring(void)
 	return MCwidgetstring;
 }
 
-const MCObjectPropertyTable *MCWidget::getpropertytable(void) const
-{
-	return &kPropertyTable;
-}
-
 bool MCWidget::visit_self(MCObjectVisitor* p_visitor)
 {
     return p_visitor -> OnWidget(this);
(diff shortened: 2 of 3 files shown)


mwieder:
Looks good as a start. Working fine here. Would like a way to expand the list beyond the hard-coded properties. Widgets can have individual properties (e.g. the isDay of the clock widget). Don't know what the mechanism would be to get the extra widget properties or to tack them onto the builtins, but it's something to think about later.

bwmilby:
I would liken the widget props more like a custompropertyset since the data can include arrays.  The 'export widget x as array y' command already does this and those two keys could conceivably be added without too much difficulty but then what is returned by 'the properties' would no longer be a set of scalar values since $state is an array (and any of the keys could also be arrays).
Making the inverse work would be a bit more complicated though.  If you merged the $state data with the fixed props then splitting them out to reconstruct a widget wouldn't be as simple.  Take the SVG icon widget - some properties can be set via 'the properties' (hilited, angle) but others don't seem to respond (iconPresetName, flip vertically).
I will also say that if you add a property that doesn't exist for a widget to the list then you get nothing.  I chased that for a bit when my first attempt didn't work.

emily-elizabeth:
Brian, is this good to go?

bwmilby:
Yes, this one should be good to go.  This will help with the `export as project` update as well.

emily-elizabeth:
Why properties returns nothing useful: It goes through a static C++ table (widgetprops[]) of 13 generic engine-layer properties (rect, visible, name, etc.). It never crosses into the LCB layer at all.

What $state does differently: It calls MCWidget::GetState(), which dispatches the LCB OnSave handler directly on the widget instance. Each widget's OnSave populates an array with all its module-level variables (segment names, selected items, SVG paths, etc.).

The fix is small and isolated. In widget.cpp, the P_PROPERTIES case (line 437) currently just falls through to the base MCControl::getprop. It could instead call GetState() and return that — or merge the two if you want the engine-layer properties alongside the LCB state. Either way it's a handful of lines in one place.

The only wrinkle: widgets without an OnSave handler (like colorswatch) would return an empty array, same as now — so no regression there.

bwmilby:
I'm not sure what you mean by "nothing useful".  Right now it returns nothing at all.  The properties that I added are useful.  You can already set them, you just can't get them.  That asymmetry is a bug in my view.

The convention is that `the properties` returns an array of scalar values.  Widget state can be anything (values can be nested arrays).  I don't think combining them works well since it breaks the convention.  If you want to add a $State key, then you would also need to add the ability to set the $State.  Export widget also gives you the $Kind key, so that would need to be there as well.

Currently, if you want to capture everything to store a widget in a non-binary form you need to manually get those 13 properties and export the widget to get the $State and $Kind.  This PR reduces that to two steps.

To restore a widget from two arrays, you just need to import the widget and then set `the properties`.

bwmilby:
I'm going to disagree about "small and isolated".  All of my changes would be needed as a base.  Adding that code to the widget.cpp file would be the wrong approach as you would be duplicating functional code.  `doGetProoerties` is where the code would need to be added to special case widgets and add state/kind.

Issue #35, opened 2026-04-05
Answer/Ask dialogs are not dark mode friendly


emily-elizabeth:
<img width="512" height="240" alt="Image" src="https://github.com/user-attachments/assets/9f9fbc66-9bc6-4d8e-bd98-9e969f64b3f1" />

emily-elizabeth:
Fixed for 0.9.8

PR #351, opened 2026-06-16
Linux appimage

emily-elizabeth:
getting the AppImage ready for distribution


bwmilby (commented):
I see the changes were reverted.  I think that the documentation templates need to have `the edition` removed since it is no longer applicable.

emily-elizabeth:
Can you guys make sure that this is working.
- VLC playback is happening both in the IDE and standalones
- IDE has a taskbar icon

mwieder:
Can indeed verify both the VLC player and the taskbar icon.
In the IDE only.
Creating a standalone failed with
1	patchelf: could not set RPATH on standalone binary — bundled VLC libs may not be found at runtime.
2	/bin/sh: 1: patchelf: not found

Note that in the IDE I am relying on the vlc libs only. I removed the app from my system.

emily-elizabeth:
What flavour of Linux are you using?

mwieder:
linux mint 22.3 64-bit (ubuntu base) cinnamon 6.6.7
linux kernel 6.17.0-35-generic

bwmilby:
I'll try.  I need to figure out what package provides the `convert` command.

ok, had to install a few things:
`imagemagick`
`patchelf`
`curl` (which was used to grab `appimage`)

It creates the file and launches fine but still has the gear icon.

bwmilby (commented):
Seems to work once dependencies installed.

After second launch, the correct icon was shown in the task bar.  In the file manager the icon was still the gear.  Standalone had gear icon both places but did build.  Able to drag a player object onto the card.  VLC crashes on my system so can't really test video.

bwmilby:
I was able to get VLC working (had to rebuild font cache) but it does not load in HXT.  If I launch the image from the terminal I do get error messages.  Here is what I get:
```
[VLC] MCVLCPlayer::MCVLCPlayer() constructing
[VLC] readlink(/proc/self/exe) -> len=43
[VLC] exe path: /tmp/.mount_HyperXmhbglC/usr/bin/HyperXTalk
[VLC] exe dir: /tmp/.mount_HyperXmhbglC/usr/bin
[VLC] probing: /tmp/.mount_HyperXmhbglC/usr/bin/vlc-plugins/plugins (exists=1)
[VLC] VLC_PLUGIN_PATH -> /tmp/.mount_HyperXmhbglC/usr/bin/vlc-plugins/plugins
[VLC] VLC_PLUGIN_PATH (env) = /tmp/.mount_HyperXmhbglC/usr/bin/vlc-plugins/plugins
[VLC] calling libvlc_new...
[VLC] libvlc_new -> FAILED (null)
[VLC] EnsureVLCInstance FAILED
[VLC] SetProperty(rect): m_rect=(34,29,237x134)
[VLC] SetProperty(rect): m_rect=(34,29,237x134)
[VLC] ~MCVLCPlayer() destructing m_view=(nil)
[VLC] ~MCVLCPlayer: complete
```
This is what I have installed:
VLC media player 3.0.20 Vetinari (revision 3.0.20-1-g2617de71b6)
Running on:
#35~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 19:30:42 UTC 2 x86_64 GNU/Linux

emily-elizabeth:
Brian, Claude asked if you can run this `VLC_VERBOSE=2 ./HyperXTalk-*.AppImage 2>&1 | head -200` as it's supposed to get the VLC messages

bwmilby:
I get the same debug output as posted above.
I just checked the plugin folder in the AppDir and it is empty, so that is likely the issue (code just checks folder, not contents)
Plugin folder also empty at source.  Using apt-get to update the package.  Rebuilt image.
And now it works once (IDE & standalone).  If video plays to end can no longer play again.  If stopped before end you can scrub and restart.  Thumb position does not track well in standalone.

emily-elizabeth:
Can you try `LD_DEBUG=libs VLC_VERBOSE=2 ./HyperXTalk-*.AppImage 2>&1 | grep -iE "vlc|error|not found|cannot|failed" | head -80`

bwmilby:
Not sure if directed at me or Mark, but here is what I get (attached).  Things work since I reinstalled VLC.  I don't think the package that I installed using the App Center brought along the plugins or at least they were not where expected.  Only real issue is the position indicator on standalone and what happens once you get to the end of a video (no additional error output).
[VLC-debug-output.txt](https://github.com/user-attachments/files/29103372/VLC-debug-output.txt)

mwieder:
OK - I rebuilt from scratch overnight and now I can build standalones with vlc from the IDE.
Not sure at all about this AppImage thing. Is this supposed to be automatic? 
I can (sort of) build .AppRun manually if I launch ./packaging/linux/build-appimage.sh build-linux-x86_64/hyperxtalk Release
but notice that the defaults don't work because a Debug build is not created for me in the specified directory.

Also running config.sh I get the error
<string>:367: SyntaxWarning: invalid escape sequence '\$'
but this is apparently ignored when building because the process continues.
I haven't yet determined where that problem is coming from.

mwieder:
@bwmilby here's my test file for the linux media player. The buttons are set up to act the way I think they should act (apropos your comment about the end of the video).
[playtest.hyperxtalk.zip](https://github.com/user-attachments/files/29107868/playtest.hyperxtalk.zip)

bwmilby:
@mwieder I'll download and try your stack.  For now, yes, you need to run the build script manually.
I can verify that the button does allow playing the video again once it completes.

@emily-elizabeth I think we need to add a linux make target that builds the app image.  `make package-linux` for example.  It can use the environment variables that are set and since we are currently mainly building release all the time, pass that.
```
package-linux-%:
	bash packaging/linux/build-appimage.sh build-linux-$*/hyperxtalk $(BUILDTYPE)

all-linux-%:
	$(MAKE) config-linux-$*
	$(MAKE) compile-linux-$*

$(addsuffix -linux,all config compile check package): %: %-$(guess_linux_arch)

```
Edited with a tested version of the end of `Makefile.linux` that works on my system.  Can't use `sh` because you will get an error for the `pipefail` option in the script.

emily-elizabeth:
Okay, added make package-linux and it made an AppImage and I was able to test on Linux Mint (it launched the IDE and that's about all I had time for) I'll do more testing tomorrow.

mwieder:
Yep. Builds and runs and makes standalones here.

I have to say for the record I'm not wild about the AppImage format for distribution. For now (0,9.X) as an end-user tool it's acceptable, but for anyone who wants/needs to make changes to stacks it's unworkable due to the fact that it runs in /tmp space and thus even experimental changes can't be saved.
As a simple example, even though I know how to get rid of that "dark,#323232,#ffffff" message in the message box, I'm forever stuck with it in the current AppImage.
Building a .deb image or an .rpm image for linux installation isn't a hard problem to solve.

mwieder (commented):
Looks good. And I think it makes sense to keep the package target separate rather than including it in the all build.

bwmilby (changes requested):
Line 44 is not needed in the makefile.  It is already the default in the base makefile.  Adding it a second time could cause confusion later on.

bwmilby (approved):

mwieder:
Good call on removing the BUILDTYPE. Normally I appreciate redundancy, but I think inheritance wins here.

PR #352, opened 2026-06-16
clear the messagebox on openstack

mwieder:
On opening the messagebox there's leftover text in the results field.
"dark,#323232,#ffffff"
It's annoying and could be easily remedied by modifying the binary stack, but it seems as easy and maybe better form to add a single line to the openStack handler to empty the field.

### DIFF
--- a/ide/Toolset/palettes/message box/behaviors/revmessageboxbehavior.livecodescript
+++ b/ide/Toolset/palettes/message box/behaviors/revmessageboxbehavior.livecodescript
@@ -98,6 +98,7 @@ on openStack
    
    __UpdateMessageColors
    __UpdateMessageFont
+	put empty into field "results"
    unlock messages
    unlock screen
 end openStack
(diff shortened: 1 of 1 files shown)


emily-elizabeth:
Does the OpenStack clear the message box every time the message box is closed and opened?

mwieder:
Well, if you close the stack you get an openStack message the next time you open it, no?

emily-elizabeth:
Personally, I think wiping out the previous results would be a bad thing.

mwieder:
That's what happens anyway. All this patch fixes is the fact that there's a string in the results field on the *first* openStack message.

emily-elizabeth:
It doesn't wipe out the previous results for me on all three platforms

mwieder:
My bad. Must have been a side effect of my testing.
OK - I'll leave it to you to fix the binary stack then to avoid a binary pull request.

PR #359, opened 2026-06-21
357-persistent-splash-screen-on-windows-11

### LINKED ISSUE #357: Persistent Splash Screen on Windows 11
> I am experiencing an issue where the splash screen in HyperXTalk (version 0.9.14) does not disappear after the application loads. It remains stuck on the screen and interferes with my workflow, as shown in the attached file.
> 
> System Details:
> 
> OS: Windows 11

Image

emily-elizabeth:
commented out the close stack "home" and have the splash screen close instead

### DIFF
--- a/ide/Toolset/home.livecodescript
+++ b/ide/Toolset/home.livecodescript
@@ -2,8 +2,8 @@
 constant kImportantTextColor = "#569e54"
 
 on startup
-//     revSetTestEnvironment($TRAVIS is true)
-
+   //     revSetTestEnvironment($TRAVIS is true)
+   
    -- MW-2011-03-13: Make sure the externals are empty on startup
    set the externals of me to empty
    
@@ -255,7 +255,22 @@ on revInternal__openStack
    
    -- Now the home stack has loaded, we can allow interrupts..
    set the allowInterrupts to true
+   
+   send "DelayClosingSplash" to me in 3 ticks
 end revInternal__openStack
+
+
+on DelayClosingSplash
+   set the cREVLaunching of stack "Home" to false
+   if there is a stack "revsplash" then
+      set the destroyStack of stack "revSplash" to true
+      close stack "revSplash"
+      if there is a stack "revSplashStackBehavior" then
+         set the destroyStack of stack "revSplashStackBehavior" to true
+         close stack "revSplashStackBehavior"
+      end if
+   end if
+end DelayClosingSplash
 
 on hxtInternal__setSplashStatus pMessage
    revInternal__Log "Message", pMessage
@@ -1726,8 +1741,7 @@ command revInternal__InitialiseFinalSteps
    if the milliseconds - sSplashStartTime < 500 then
       wait 500 - (the milliseconds - sSplashStartTime) milliseconds
    end if
-   --   close stack "home"  // HXT
-   close stack "revSplash"
+   close stack "home"  // HXT
    
    revInternal__Log "Message", "Resetting Cursors"
    if the platform is "MacOS" then
--- a/ide/Toolset/home.livecodescript
+++ b/ide/Toolset/home.livecodescript
@@ -2,8 +2,8 @@
 constant kImportantTextColor = "#569e54"
 
 on startup
-   //     revSetTestEnvironment($TRAVIS is true)
-   
+//     revSetTestEnvironment($TRAVIS is true)
+
    -- MW-2011-03-13: Make sure the externals are empty on startup
    set the externals of me to empty
    
@@ -105,7 +105,6 @@ end openStack
 
 // HXT
 on CloseStack
-   revInternal__Log "log", "Close Stack"
    close file sLogPath
 end CloseStack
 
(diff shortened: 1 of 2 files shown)


bwmilby:
Line 230 should probably move...
```
         hxtInternal__setSplashStatus ""
```
This is blanking the status before the announced work in the previous line is complete.  It would be helpful to know if the hang was happening before or after that work.

I would move it to immediately before current line 251 (my version)
```
   revInternal__Log "Message", "Closing Splash (for real this time)"
   hxtInternal__setSplashStatus ""
   close stack "revSplash"
```
I'm also a little suspicious of line 1729 (`close stack "home"`) - I don't think it belongs there.  The home stack can't really close either but issuing that command seems to skip everything after it.  I'm not sure why the splash would even be closed in that location.  With that line there, the log ends at "closing splash."  When I comment out that line, then I do get the next few messages including the one for leaving final initialization.  But then, I do not get the two messages after closing the splash stack.  As a test, I added a message before closing splash (included above) and that did make it to the log.  Here is the end of my log when launching from Mac-bin after moving the line and commenting out 1729:
```
      Closing Splash
      Resetting Cursors
      Initialising revFrontScript
    Leave Final Initialisation
    Closing Splash (for real this time)
```
(I'd also get rid of the first "closing splash" message.  Leave in the delay which just ensures the splash is visible for half a second.)

emily-elizabeth:
So I commented out the `close stack "home"` and added the deferred close stack right after the `close stack "revSplash"`

bwmilby:
I decided to take a look at LC11 just to see if anything had really changed since LC9.  The `close stack "home"` call was still there in the same spot (and I really am not sure why it works).  It looks like the open/close file got commented out at some point in the HXT code but I think it probably should still be used that way.  Something somewhere else must have changed that is causing some of the later log messages to get lost or just not sent to the file.

Looking at the code, we are losing a chunk of the log due to how it is being managed.  The method used to manage the log file looks to be causing other issues.  If you need it fresh each launch, it would be better to rotate or delete the old file than open the file and leave it open (e.g. `delete file sLogPath` right before the "startup started" log message).  Just changing back to the original code allowed the last few steps to make it to the log file that were also being lost.

Compare (starting at final steps, original method):
```
    Performing Final Steps...
      Enter Final Initialisation
      Installing Remaining Libraries
      Setting up verbose debugging
      Closing Splash
      Resetting Cursors
      Initialising revFrontScript
    Leave Final Initialisation
  Leave revInternal__openStack
  -------- Startup Ended: Mon, 22 Jun 2026 21:04:26 -0400
```
to:
```
    Performing Final Steps...
      Enter Final Initialisation
      Installing Remaining Libraries
      Setting up verbose debugging
      Closing Splash
```
Also on the startup side, compare (first 5 lines, original method):
```
-------- Startup Started: Mon, 22 Jun 2026 21:04:23 -0400
  Enter Startup
    Enter Path Setup
    tBuildDir = build-mac-arm64
    tPackageDir = mac-arm64-bin
```
to
```
  Enter Pre-Open Stack
    Enter Defaults Initialisation
    Trying to set the default text of stack home
  Leave Defaults Initialisation
Leave Pre-Open Stack
```
The current log method is missing 187 lines from what was captured done the original way.

emily-elizabeth:
When the splash screen is closed, the CloseStack message is fired in the `home` stack and it closes the log file.

bwmilby:
The issue isn't on the close side.  There is a significant part of the log that is missing due to the way the file is being managed.  You are missing everything that happens at startup before you get to preOpenStack.  The better method is to reset/delete the file manually and use the previous method to open/close the file for every write.

I will add that all of the log gets saved - even with that call to close the home stack that is way too early in my opinion - using the original code for logging.

Also, a good reason for the original process (open/close file for each write) is to ensure that the log is not corrupted in the event of a crash during startup.  With a clean open/close the file is always guaranteed to remain in a good state.

emily-elizabeth:
From the tester:

> I installed the latest HyperXTalk installer you sent me, and I’m happy
> to confirm that the “Persistent Splash Screen on Windows 11” issue has
> been resolved. Thank you for your hard work.

bwmilby:
I think it would be good to narrow down the actual fix.  You made 2 significant changes.  I'm inclined to think that the commented out code is more likely to have resolved the issue than the delay close addition but that is just a guess.  If the delay close was in the previous update sent to the user that did not show improvement, then it is very likely not a needed change.

Relocating the `send ... in 3 ticks` has no impact.  What that statement does is call that code immediately after the current handler exits.  Changing where it is located would only make a difference if the handler was exiting early before the send call.

emily-elizabeth:
Sent the test build to a couple others who had the same issue so see what they say.

bwmilby:
My concern is that we have not identified what was causing the issue.  Two completely different changes were implemented.  The send it time should not be added if not needed.

You commented out the code that would establish the file associations on windows.  I'm not convinced that getting rid of that code is the best solution.  I would highly suggest that we return the logging to the original state (open/close on every add), add some logging around that section of code to see what is actually happening, and then figure out the way forward.  The way the log file is currently being handled ensures we are missing log entries.  Look at my examples above.  A large section of startup log and the final pieces at the end are not logged so you do not know where the handler actually stopped working.

bwmilby:
After restoring log process, here is an example addition to figure out where things are stopping
```
   revInternal__Log "Message", "Win file association check"
   if $REV_TOOLS_PATH is empty and (revInternal__CheckFileAssociations() is true) and (the platform is "win32") then
      revInternal__Log "Message", "Win32, tool path empty, check file assoc true"
      if revIDEGetPreference("cDontShowFileAssociationsDialog") is not true then
         revInternal__Log "Message", "show modal"
         modal stack "revFileAssociations"
         revInternal__Log "Message", "after modal"
      end if
   end if
   revInternal__Log "Message", "after Win file association check"
```

bwmilby:
I'm going to suggest that everything be reverted and then this single change made:
```
modal stack "revFileAssociations.livecodescript"
```

bwmilby:
I think I found the actual problem.
`ide/Toolset/palettes/revfileassociations.livecodescript`
Line 1 is incorrect...
`script "revfileassociations.livecodescript"`
This is why the stack could not be found, the stack short name includes the extension.  The `.livecodescript` needs to be removed and that should fix the issue without needing to touch `home.livecodescript` at all.

emily-elizabeth:
Restored home stack to the version with the log fixes and corrected the short name of the revfileassociations stack. Are you able to test to ensure that the changes fix the issue?

bwmilby:
This does prevent the hang, but that dialog does not seem to save the "don't show again" preference.

- Launched Win11 25H2 VM
- Installed HXT 0.9.14
- Opened HXT, verified issue present, closed HXT
- Replaced revfileassociations.livecodescript with downloaded version from PR
- Opened HXT, presented with dialog to map associations (chose no), closed HXT
- Opened HXT, checked "don't show again" on association dialog, closed HXT
- Opened HXT, dialog still appeared

While this is a separate bug from the original report, it is still not "right".  When looking at the original repo, this was not a livecodescript file.  I think we may want to de-scriptify this and revert to the original file.  The other option is to figure out why the scriptified version of the dialog isn't working properly.

I replaced the file with the revfileassociations.livecode file from the 9.6.3 archived repo and it functioned properly (although still branded LiveCode).  I suggest we go back to the .livecode file and rebrand it for this PR.

emily-elizabeth:
Seems the extensions weren't being registered with the installer, so I added them. I wonder why it never triggered on the development machine as LC was never installed.

emily-elizabeth:
Instead of checking `word 2`, we need to check `trueWord 2` as it's including the quotes in word 2

bwmilby:
That is a good catch (trueWord).  When I was looking at it the other night I wasn't in troubleshooting mode and completely missed that.

PR #360, opened 2026-06-23
Prettyprint json

mwieder:
Added the ability to prettyprint exported json through JSONExportPretty. Would love to condense this down to a single invocation of JSONExport, but the LCB "optional" qualifier is parsed correctly but nonfunctional. Someday maybe dig down into the LCB code and fix this.


bwmilby:
To clarify, optional just means that the variable can be `nothing` or a value (`optional any` can be nothing or a value of any type).  Currently LCB handlers must have the correct number of parameters used.

I suggest untracking the api.lcdoc file.  It is only getting added since it is tracked already and shouldn't be in the repo.

mwieder:
OK. Wasn't sure where in the build process the lcdoc file was generated. Untracked now.
The lack of parity between LCB and LCS is annoying.

bwmilby:
There is a PR out there to fix the docs build process but the LCB files are processed on every platform currently and compiling will build those lcdoc files.  It is only the main dictionary and guides that need to be assembled for packaging separately.  It can be done by launching the IDE or using `make` with the PR change that is posted (integrated the process LC used to build docs).

mwieder:
OK - pushed a new version to accomplish separation of tasks.
The heavy lifting is still done by the lcb library for speed but the flexibility of modifying the spacing and format is now moved to the backscript library. This because lcb isn't (yet?) up to the task of optional arguments.

put JSONExport( value, prettyprint?, spacing, expand?)
Where
value is a HyperXTalk expression
prettyprint? is a boolean true to prettyprint the JSON (default=false)
spacing is the number of spaces to indent each level (default=4)
expand? if true will place "{" and "[" separators on new lines (default=true)

bwmilby:
Only issue I can see is in built apps - backscript library would need to be included if JSONExport command is used (or that handler would need to included somehow).

mwieder:
That's a good point. I originally put that handler into the home script but ran into errors.
Hmmm. Looking at revsblibrary I'm thinking maybe the way around that is to create a new entry in script-libraries to handle JSONExport rather than having it in an IDE script.

bwmilby:
Yeah, just have to ensure inclusions work for it.  An even better approach would be to add frontscript/backscript support to LCB projects.  That would allow this kind of addition to naturally travel with the code that depends on/uses it.

mwieder:
The current front and back scripts handle *mostly* IDE tasks, so there's not much point in adding the extra bulk to standalone apps. I suppose we could create different front/back scripts for standalones, but for the case of JSON exports I think it makes more sense to just add it to the possible inclusions list. Not every standalone is going to need JSON export capability, and I like the idea of trimming the fat from standalone apps when feasible.

mwieder:
OK - pushed a new commit that handles standalones.
Added a script library for the export command, moved it from the backscript.
It now loads as a backscript and can be selected as an inclusion for standalone building.

bwmilby:
LC handled this differently in LC 11.  They built a separate command for a prettyPrint version of the export (`jsonEncodeForDisplay`) that still took a single argument.  I wonder if that may be a cleaner option?  You could have a getter/setter for the 2 options (spaces and braces) if someone didn't want to use the default behavior.

mwieder:
I haven't seen LC11, so I'm taking a clean-room approach. Interesting idea, but how do getters/setters work with libraries? I'm not wild about the extra backscript for one function, but I don't see how to do this without making a system-level property, which seems equally cumbersome.

bwmilby:
Here's an example property from the iconsvg library:

```
private variable sCurrentFamily as String

<snip>

public handler getCurrentIconFamily() returns String
	return sCurrentFamily
end handler

public handler setCurrentIconFamily(in pFamily as String) returns nothing
	if pFamily is not among the keys of sFontArray then
		throw "Icon family '" & pFamily & "' does not exist."
	end if
	
	put pFamily into sCurrentFamily
end handler
```

Would just need to create 2 library properties that are set to reasonable defaults but allow customization by the user if desired.

mwieder:
OK - so not actual properties, just more handler calls. Have to say I'm not wild about that approach. Prettyprinting JSON objects would get unwieldy and awkward.

Exporting JSON from HyperXTalk objects would continue as at present:
JSONExport pObject

So there would be a minimum of two calls to prettyprint JSON:
setTheJSONPrettyPrintTo true #default is false
JSONExport pObject

and a maximum of four handler calls:
setTheJSONPrettyPrintTo true #default is false
setTheJSONPrettyPrintSpacingTo 3 #default is 4
setTheJSONPrettyPrintBraceExpansionTo true #default is false
JSONExport pObject

That seems like a lot of work to go through when compared with optional handler arguments:
JSONExport pObject, true, 3, true

bwmilby:
They are library properties - you can set them once and they persist for the session.  If you want to modify the way pretty print works, you just make those settings once.  A combined setter could also work.  I would use separate handlers for regular and prettyPrint.  In the prior convention, you would have to add an argument to get pretty so rather than add an argument you just change the handler name.
```
JsonExport pObject

ConfigureJsonPrettyPrint pSpacing, pExpansion
   /* effectively
   setTheJSONPrettyPrintSpacingTo pSpacing
   setTheJSONPrettyPrintBraceExpansionTo pExpansion
   */
JsonExportPretty pObject
```
To be honest, I'm not sure there is a speed advantage to export via LCB vice LCS.  If we put JSON in the engine and called both ways (import/export) using C(++) then there would be an advantage for sure.  As it stands you end up doing quite a bit of work in LCB to even pass to the yyjson code so it probably isn't worth it for export.

mwieder:
Re: speed I think if you're wanting to prettyprint the JSON that's a human-facing end result so timing probably isn't critical. Exporting HxT objects to JSON without prettyprinting internally could still have a speed advantage. My guess is that the speed difference IRL would end up being minimal.

In terms of coder experience, though, I still prefer the single function call to export JSON with optional arguments rather than the multiple step approach. It's only one handler to remember and one dictionary entry if you need to look up how to set non-default options.

bwmilby:
I just looked at LC11 and it has no options.  It is either for display or not.  They do have encode as ascii and as utf8 though (each is a separate handler).

One thing we could consider is maintaining script compatibility with LC 11.  That way if someone created a stack with LC 11 and we opened it in HXT it would still work.  I would handle this the way PhotonJSON does by just adding the LC handler names to the livecodescript file that would redirect into the ones for HXT.  They did change the names in 11 and deprecated the old ones.  mergJSON is still there though (but not sure why since the new stuff is better).

emily-elizabeth:
What's the status on this? Is it good for me to test and merge?

mwieder:
Since it's my PR I shouldn't give the OK on this. I'll say that it's working as is and maybe we could augment it later by adding wrapping functions to give new names. I don't know what the new names are, so I'll leave that to someone else. @bwmilby : any comments?

bwmilby:
I think it is good for test now.  As Mark mentioned the addition of compatibility mappings should be separate anyway.  If we later discover a better way to make the LCS code available we can shift at that time.  This change will be necessary to working on the project export code.

emily-elizabeth (approved):

emily-elizabeth:
I can't get this to work. Keep getting `button "Button": execution error at line 10 (vectordot: error in first parameter), char 8` when running the following code on macOS HyperXTalk 0.9.15:

```
on MouseUp
   local tData
   
   put "Alice" into tData["name"]
   put 30 into tData["age"]
   put "Toronto" into tData["city"]
   put true into tData["active"]
   
   local tJSON
   put JSONExport(tData, TRUE, 4, TRUE) into tJSON
   put tJSON
end MouseUp
```

mwieder:
It's still working for me. Open the Project Browser and make sure that com.hyperxtalk.library.jsonexport is being loaded.

bwmilby:
I just checked out the PR branch, did a `make config-mac` and `make compile-mac` and then `hxide` (alias to the IDE in the mac-bin folder).  I created a stack and pasted your code into a new button script.  Message box gave me the expected output:
```
{
    "active": true,
    "city": "Toronto",
    "age": "30",
    "name": "Alice"
}
```

emily-elizabeth:
I'll give it another whirl tomorrow (Monday) as I'd like to add it to 0.9.15

bwmilby:
Looks like I was a commit behind.  I pulled again and am seeing the same error that Emily reported.

bwmilby:
I did a bit of looking and found that the LiveCode instance of the LCB library was having around in the tree in several places.  Once I removed that, I got rid of the vectordot error but for some reason the script library isn't loading.  I manually loaded it and it worked, but didn't have time to figure out why it wasn't loading.

After some additional troubleshooting I'm pretty sure it is due to more old artifacts in the build tree.  I ended up doing essentially a clean build (deleted `_build`, `_cache`, `build-mac`, and `mac-bin` directories) and things worked correctly.

I also built a standalone for Mac with manually selected inclusions and it worked properly.

Documentation needs some corrections to build properly.  All changes in the new script library file:
- "?" is not a valid character in a parameter name for the dictionary parser.  Rename parameters or we will need to look into changing the parser.
- "Example" vice "Examples"
- In the two blocks of JSON, must quote the word before the ":" otherwise it confuses the parser and does not render correctly.
- I believe that all of the code in examples should compile (I saw a test somewhere that checks).  If using that standard, then all non-code lines should be commented.  Doing this takes care of my above comment as well.

mwieder:
OK - made the changes to the documentation and removed the '?'s. I wouldn't at all mind updating the docs parser to allow them, as I use them quite extensively for boolean arguments. At any rate, give this a try.

mwieder:
Any chance the build artifacts are getting confused because the directory is named 'json' just like the lcb extension folder? The script stack itself has a unique name, but maybe it's worth renaming the enclosing folders.

mwieder:
All right - I have no idea what's going on now. I rebuilt completely from scratch, and now JSONExport still works as before but JSONImport throws an error of "can't load foreign library". So calls into the JSON lcb library are working when issued from the script library but not from a stack script.

bwmilby:
Just validated that docs build completely with the changes.

bwmilby:
`JSONImport` uses the `yyjson` compiled code.  Is that not getting built properly?  It builds correctly on my Mac.  I just started a clean build on my Linux seat, but probably won't see the result tonight.

bwmilby:
Compile finished overnight and this morning I was able to test import and export successfully.

emily-elizabeth:
Tested on macOS 26.5.2 and it seems to be working correctly.

emily-elizabeth:
Tested on macOS 15.7.7 and Windows 11

emily-elizabeth:
Tested on Ubuntu 25.04
Seems like everything is good with this now, time to merge! 🙂

mwieder:
I'm still getting the 'can't load foreign library' error on JSONImport. I verified that yyjson.o is being built correctly and rebuilt and reinstalled the lcb extension several times, and rebuilt completely from scratch a few times, but the error persists. Annoyingly there's also an empty com.livecode.library.json folder that gets rebuilt by gyp.

mwieder:
Trying to unravel this thing. The error is apparently thrown because of the "use com.livecode.foreign" line in the lcb. Don't know why, but it seems to be the only "com.hyperxtalk..." library that uses the foreign library. Other com.livecode lcb stacks that call on that library work fine.

The build process also creates several empty packaged extensions folders:
com.livecode.library.animationengine
com.livecode.library.hxtpm
com.livecode.library.json
com.livecode.library.macscrollenhancement2
com.livecode.library.shakecontrol

@bwmilby are you saying that JSONImport works for you with a current linux build?

bwmilby:
Yes.  I did a make clean, config, build and it worked properly.

I also see the empty folders.

I'll add that I'm using an alias to launch the IDE from the linux*bin folder via the terminal.

Also downloaded the appimage and tested successfully.

mwieder:
OK - I git cloned a new repo from Emily's main branch, cleaned it, and built it anew. Still get the 'foreign' error when trying a JSONImport. So I can eliminate anything in my repo that might be amiss. The error that's being thrown is in the lcb file at the line
use com.livecode.foreign
Possible glibc difference? Any other ideas?

PR #370, opened 2026-06-30
Feature/linux hidpi

### LINKED ISSUE #356: Linux: IDE UI does not scale on HiDPI displays
The HyperXTalk IDE renders at 1x on HiDPI Linux displays regardless of the system scale factor. This affects Wayland sessions with fractional scaling (e.g. 125%) where the IDE appears tiny compared to the rest of the OS. GTK scaling environment variables and GNOME's text-scaling-factor have no effect because the IDE UI is drawn by the engine's own rendering system rather than GTK widgets. Fixing this requires engine-level support for reading and applying the display scale factor.

emily-elizabeth:
Linux HiDPI infrastructure (GTK3/GDK)
Lays the groundwork for HiDPI support on Linux using GDK 3.22+ monitor APIs. The detection and notification infrastructure is fully wired up; pixel-scaled rendering is disabled pending a deeper change to MCLinuxStackSurface.
What's implemented:

MCLinuxGetLogicalToScreenScale() — reads the primary monitor's GDK scale factor, safe to call before MCScreenDC::open() (uses gdk_display_get_default() rather than the member dpy)
device_getdisplays now populates MCDisplay::pixel_scale per monitor via gdk_monitor_get_scale_factor()
All coordinate conversion functions (logicaltoscreenpoint/rect, screentologicalpoint/rect) implemented, mirroring the Windows w32dc.cpp pattern
Runtime monitor scale-change signals: notify::scale-factor connected per-monitor in MCScreenDC::open(), plus monitor-added to catch hot-plugged displays
MCResPlatformInitPixelScaling, MCResPlatformGetDefaultPixelScale, MCResPlatformHandleScaleChange, and the remaining capability stubs all implemented

Why rendering isn't enabled yet:

MCLinuxStackSurface creates its pixbuf at logical pixel dimensions and blits via MCX11PutImage, bypassing GDK's scaling layer. XWayland already handles logical→physical scaling at the compositor level, so setting view_setbackingscale > 1 causes the tilecache to render at N× into a 1× raster, garbling the display. Both MCResPlatformSupportsPixelScaling() and the view_setbackingscale call in onexpose() are disabled with TODO [[ HiDPI ]] comments marking what needs to change once MCLinuxStackSurface allocates at physical pixel dimensions.
Tested: Renders correctly at 1×, 2×, and fractional GDK scale factors — XWayland compositor scaling handles all cases transparently.

### DIFF
--- a/engine/src/lnxdc.cpp
+++ b/engine/src/lnxdc.cpp
@@ -467,13 +467,18 @@ void MCResPlatformInitPixelScaling(void)
 }
 
 // GTK3 (GDK 3.22+) exposes per-monitor scale factors via
-// gdk_monitor_get_scale_factor(), so HiDPI is fully supported.
-// GDK handles DPI awareness automatically (no process-level flag like
-// Windows SetProcessDPIAware), so we return true unconditionally —
-// matching the macOS desktop-dc.cpp implementation.
+// gdk_monitor_get_scale_factor(), so HiDPI detection infrastructure is
+// in place.  However, MCLinuxStackSurface currently creates its pixbuf
+// at *logical* pixel dimensions and blits via MCX11PutImage, bypassing
+// GDK's scaling layer.  XWayland already handles logical→physical scaling
+// at the compositor level, so enabling pixel scaling here causes the
+// tilecache to render at Nx into a 1× raster, garbling the display.
+//
+// TODO [[ HiDPI ]]: return true once MCLinuxStackSurface creates surfaces
+// at physical pixel dimensions (and coordinate math is updated accordingly).
 bool MCResPlatformSupportsPixelScaling(void)
 {
-    return true;
+    return false;
 }
 
 // Scale factor is read from GDK at runtime; it cannot be set by the user
--- a/engine/src/lnxdc.cpp
+++ b/engine/src/lnxdc.cpp
@@ -52,7 +52,6 @@ along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
 #include "font.h"
 #include "redraw.h"
 #include "resolution.h"
-#include "mode.h"
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -469,9 +468,12 @@ void MCResPlatformInitPixelScaling(void)
 
 // GTK3 (GDK 3.22+) exposes per-monitor scale factors via
 // gdk_monitor_get_scale_factor(), so HiDPI is fully supported.
+// GDK handles DPI awareness automatically (no process-level flag like
+// Windows SetProcessDPIAware), so we return true unconditionally —
+// matching the macOS desktop-dc.cpp implementation.
 bool MCResPlatformSupportsPixelScaling(void)
 {
-    return MCModeGetPixelScalingEnabled();
+    return true;
 }
 
 // Scale factor is read from GDK at runtime; it cannot be set by the user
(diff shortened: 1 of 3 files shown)


engine/src/lnxdc.cpp
```diff
@@ -308,13 +310,11 @@ bool MCScreenDC::apply_partial_struts(MCDisplay *p_displays, uint32_t p_display_
...
 }
 
-// IM-2014-01-29: [[ HiDPI ]] Placeholder method for Linux HiDPI support
 bool MCScreenDC::platform_getdisplays(bool p_effective, MCDisplay *&r_displays, uint32_t &r_display_count)
 {
 	return device_getdisplays(p_effective, r_displays, r_display_count);
 }
 
-// IM-2014-01-29: [[ HiDPI ]] Refactored to handle display info caching in MCUIDC superclass
 bool MCScreenDC::device_getdisplays(bool p_effective, MCDisplay * &r_displays, uint32_t &r_display_count)
```
mwieder: Why are we passing p_effective and then ignoring it?

emily-elizabeth:
Tested on Ubuntu, Linux Mint, Debian and Fedora

engine/src/lnxdc.cpp
```diff
@@ -384,28 +390,53 @@ MCPrinter *MCScreenDC::createprinter(void)
...
+    if (!MCResGetUsePixelScaling())
+        return 1.0;
+
+    // Use the default GDK display rather than MCScreenDC::dpy so this free
+    // function can be called before MCScreenDC::open() (MCResInitPixelScaling
+    // is called twice in globals.cpp: once before and once after open()).
+    // gdk_display_get_default() returns NULL until gdk_display_open() runs,
+    // in which case we fall back to 1.0 — the second post-open call then
+    // picks up the real monitor scale.
+    GdkDisplay *t_display = gdk_display_get_default();
```
mwieder, line 424: If I understand the comment above correctly, should the call to gdk_display_get_default() be in a while loop waiting for a not_NULL response?

mwieder (commented):
I don't have one of those fancy monitors to test this physically, but I reviewed the code, compiled and ran it, and I can attest that it doesn't break anything.

emily-elizabeth:
> I don't have one of those fancy monitors to test this physically, but I reviewed the code, compiled and ran it, and I can attest that it doesn't break anything.

I don't have a fancy monitor either, but I did the scaling to simulate the HiDPI and when I tried it on Fedora it didn't go all "tiny"

engine/src/lnxdc.cpp
emily-elizabeth: Windows does the same — p_effective is never used inside any platform implementation. The intent is clear from uidc.cpp: p_effective drives cache invalidation at the superclass level only; both viewport and workarea are always populated in the returned MCDisplay array and callers pick the appropriate field. Our device_getdisplays should just mark the parameter unused:

engine/src/lnxdc.cpp
emily-elizabeth: A spin-wait would actually be wrong here. gdk_display_get_default() won't become non-NULL on its own — it only gets set when gdk_display_open() (or gtk_init()) runs. Looping before that happens would block forever. The NULL case is intentional: the first call is a "not ready yet, return a safe default" situation, not a "keep retrying" one. The real value comes from the second call after open().
The comment is just poorly worded — it reads like a known deficiency rather than an intentional two-phase design.

emily-elizabeth:
Can you see if the changes made put the splash screen on the right monitor?

mwieder:
Nope. Still the same behavior.

emily-elizabeth:
And this was working before the HiDPI changes?

mwieder:
I think it changed with the GTK2->GTK3 upgrade. I'm not too concerned about the splash screen since things otherwise work, just noting the change.

PR #372, opened 2026-07-01
Feature/linux gtkpopover

emily-elizabeth:
Linux: Native GtkPopover for WM_POPOVER windows
Implements native GTK3 popover look and feel for stacks opened with WM_POPOVER on Linux, replacing the previous stub behaviour. Also removes leftover DnD and poll debug output.
What's new
WM_POPOVER now renders as a native GtkPopover with the platform arrow, rounded corners, and drop shadow. The implementation uses a proxy GtkWindow (override-redirect, full-screen, bound to the engine's MCdpy display) to satisfy GtkPopover's requirement for a relative_to widget with a realized parent window. A GtkFixed inside the proxy acts as the anchor widget; its coordinate space equals screen coordinates, so MCpopoveranchor can be passed directly to gtk_popover_set_pointing_to(). Stack content is rendered into a GtkDrawingArea inside the popover via the engine's existing surface/pixbuf pipeline.
Files changed

engine/src/lnxstack.cpp — full WM_POPOVER branch in realize() and platform_openwindow(); MCLinuxPopoverClose() and MCLinuxPopoverGetGdkWindow() helpers; Unlock() pixbuf bridge; on_popover_da_draw() bootstrap and CAIRO_FORMAT_RGB24 blit.
engine/src/lnxdcs.cpp — closewindow() and destroywindow() branches for WM_POPOVER teardown.
engine/src/lnxdclnx.cpp — GDK_BUTTON_PRESS outside-click dismiss via GdkWindow parent-chain walk.
engine/src/lnxdnd.cpp, engine/src/dispatch.cpp — removed leftover debug fprintf output.

**Bugs fixed during development**

Blank popover content — GTK3 converts expose events to "draw" signals before the engine's event loop sees them, so onexpose() was never called. Fixed by bootstrapping a render inside on_popover_da_draw() when no pixbuf exists yet.
Content rendered transparent — LockPixels() returns kMCGRasterFormat_xRGB, leaving the alpha channel at zero. Passing the pixbuf through gdk_cairo_set_source_pixbuf() premultiplied every pixel to nothing. Fixed by wrapping the raw pixel data in a CAIRO_FORMAT_RGB24 Cairo surface instead, which treats all pixels as fully opaque and matches the engine's byte order.

Content never drawn at all — gtk_widget_show_all(proxy) doesn't recurse into GtkPopover (it's a popup, not a container child), so the GtkDrawingArea's VISIBLE flag was never set and it was never mapped. Fixed by calling gtk_widget_show(da) explicitly before gtk_popover_popup().

Popover closed on every inside click — modal=TRUE causes GTK to install its own button-press-event handler that calls gtk_popover_popdown() whenever event->window is anything other than the popover's own GdkWindow — including the GtkDrawingArea. This fired before our code ever ran. Fixed by setting modal=FALSE and managing dismissal ourselves: the full-screen override-redirect proxy intercepts outside clicks, and lnxdclnx.cpp walks the GdkWindow parent chain to decide whether to call wclose().

Inside clicks not reaching content — The proxy was raised last, putting it above the popover in the X11 stacking order. Without a modal grab, all clicks landed on the proxy instead of the DA. Fixed by raising the popover again immediately after the proxy, giving a final order of popover > proxy > main window.

Content still not interactive — GtkDrawingArea requests no GDK input events by default. Without an explicit event mask, X11 delivered button press and motion events to the nearest ancestor that had them (the popover's GdkWindow), which the dispatcher didn't know about. Fixed by calling gtk_widget_add_events() with the full pointer and keyboard mask before realization.

Crash on mouse enter — The engine opens its own GDK display connection via gdk_display_open() (MCdpy), separate from GTK's gdk_display_get_default(). Engine cursors are created on MCdpy; the proxy window was created on the default display. GDK asserts display equality in gdk_window_set_cursor_internal() and aborted. Fixed by calling gtk_window_set_screen(proxy, gdk_display_get_default_screen(MCdpy)) before realization so all descendent GdkWindows share the engine's display.


mwieder:
I'm occasionally (maybe 1/3 of the tries) getting a fatal X-window error.
Seems to be coming from a call to x11::XGetWindowProperty at line 229 in lnxdc.cpp where
i=12
dpy = (GdkDisplay *) 0x555556bfe0e0
This is in apply_partial_struts where
t_displays=(MCDisplay *) 0x55555da6b490
t_monitor_count=3

mwieder:
Also... when it does run the popover window appears and I get the proper close message in the messagebox, but the contents of the popover window are invisible. I can tell there's a field there because the cursor changes, but it seems the characters I type are as invisible as the field.

emily-elizabeth:
> Also... when it does run the popover window appears and I get the proper close message in the messagebox, but the contents of the popover window are invisible. I can tell there's a field there because the cursor changes, but it seems the characters I type are as invisible as the field.

I managed to get this once in my testing on Ubuntu. See if the changes help both of the issues

mwieder:
The new changes seem to have solved the fatal error problem. The popover window controls are still invisible.

mwieder:
Seems like a complicated fix, but unfortunately didn't make a difference for me.
Possible data point here is that if I change the popup stack background color while it's open and then click to close it I get the 'save stack?" message and the pop is temporarily visible.
afterPropertyChange

emily-elizabeth:
That's due to the fact that you made changes to the stack, so normal behaviour for HyperXTalk. Use the `cantModify` property to make changes but not save them (or be prompted to save them)

mwieder:
Yeah, just pointing out the fact that I can verify the controls on the popover stack are present, they're just invisible.

emily-elizabeth:
Okay, made some changes and tested on Ubuntu, Linux Mint, Debian 13 and Fedora. Can you give it a test at your earliest convenience?

mwieder:
Yep. Looks good now.

PR #374, opened 2026-07-04
Got tired of builds taking too long.

mwieder:
Avoid unnecessary downloads.
 Don't rebuild files unless we grabbed new versions.
 Still always downloads libpng. Don't have a good way to determine version.
 But rebuilding on linux is much faster now, especially due to not having to rebuild CEF and openssl unless necessary.


emily-elizabeth:
Tested and it still compiles on Linux, Window and macOS

emily-elizabeth (approved):

PR #377, opened 2026-07-08
Fix backdrop to span all connected monitors

### LINKED ISSUE #376: backdrop across multiple monitors
> Maybe adding the ability to have a backdrop across multiple monitors, since setups with multiple monitors are common these days.

A request from the survey feedback.

emily-elizabeth:
The backdrop window was previously sized to the primary screen only (MCPlatformGetScreenViewport(0, ...)), leaving secondary monitors uncovered.

enablebackdrop() in desktop-dc.cpp now iterates all connected screens via MCPlatformGetScreenCount() and MCPlatformGetScreenViewport(), computing the union rect of all screen bounds before sizing the backdrop window. This ensures the backdrop covers every monitor in any arrangement (side-by-side, stacked, mixed positive/negative virtual coordinates).

Single-monitor setups are unaffected.

Fixes #376


emily-elizabeth:
This is fixed for macOS. I don't have multiple monitors on the Windows or Linux laptop.

emily-elizabeth:
If you don't have multiple monitors I guess there is not too much we can do about that.

bwmilby:
I'll see if I can hook an additional monitor to my Linux laptop (it has outputs, just have not tried yet).  I have not configured my Win11 VM to build yet - I'll see if I can get a second display though.

emily-elizabeth:
Oh, I know there's no code for Windows and Linux

bwmilby:
Ok, misinterpreted the review request.  I can build and do a double check.

emily-elizabeth:
Thanks. I like to make sure someone else sees the same thing I see 😉

bwmilby (approved):
Appears to work correctly on my computer with one additional monitor attached.

emily-elizabeth:
Should we ship it with just macOS support, or should I look at getting an external monitor for the Windows/Linux laptops?

emily-elizabeth:
I ordered a second monitor for the Windows and Linux laptops, so we'll wait and have the feature on all platforms

bwmilby:
If the request came from a Mac user, I wouldn't see an issue with rolling it out for Mac first.  Would be good to rebase it on main if going to wait though.  Right now the branch still includes revbrowser due to when it was started.

emily-elizabeth:
Still going to be a while for 0.9.16

emily-elizabeth:
External monitor is supposed to be delivered tomorrow

emily-elizabeth:
updated backdrop on Windows to handle multiple monitors

emily-elizabeth:
So doing the "window-per-monitor" way of doing things, like macOS and Windows, causes an issue where the main monitor does not have a backdrop. Spanning a backdrop window across all available monitors but leaves a section of the backdrop missing on the secondary monitor, this is due to the window manager cutting out parts for the panels. This was only tested on Ubuntu, so far and further testing on other distros is going to take place.

emily-elizabeth:
Issues on debian and Fedora, testing will continue tomorrow or Monday

emily-elizabeth:
*yay* Finally got this working on Ubuntu 25.04, just need to test on the other distros

emily-elizabeth:
Alrighty, this is finally ready for review if anyone has extra monitors for their Windows and Linux boxes

mwieder:
s_wm_is_muffin is undefined.
Should that be s_wm_is_marco ?

emily-elizabeth:
> s_wm_is_muffin is undefined. Should that be s_wm_is_marco ?

Is that at line 61 of lnxdclnx?

emily-elizabeth:
> s_wm_is_muffin is undefined. Should that be s_wm_is_marco ?

Issue has been fixed

emily-elizabeth:
On Linux, there is an issue where other windows cannot be brought in front of HyperXTalk when the backdrop is in use. There is another branch to deal with this

PR #379, opened 2026-07-09
removed the forecolor of the answer and ask stacks

emily-elizabeth:
Hopefully to get the stacks to work properly in dark mode


bwmilby:
Here is what actually changed in the stacks:
https://github.com/Milby-Computer/hxt-bst/compare/main...fix/answer-ask-dialogs-darkmode

PR #385, opened 2026-07-12
arm64 mac: call applyscroll() in setgeom() so menubar group scrolls out of view

### LINKED ISSUE #384: macOS menu group is not hidden
Image

emily-elizabeth:
The osxstack.cpp setgeom() calls applyscroll() before adjusting the viewport, but the arm64 stub in mac-stubs-arm64.mm was missing that call. As a result, setting the menubar property on Apple Silicon never triggered the scroll that pushes the menu group above the top of the card.

### DIFF
--- a/engine/src/mac-stubs-arm64.mm
+++ b/engine/src/mac-stubs-arm64.mm
@@ -1733,6 +1733,8 @@ void MCStack::setgeom()
         return;
     }
 
+    applyscroll();
+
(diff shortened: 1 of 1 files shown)


emily-elizabeth:
image

emily-elizabeth:
Tested on macOS 15.7.7, macOS 26.5.2, Windows 11 and Ubuntu 25.04

PR #387, opened 2026-07-12
New error code generation (pr315a)

bwmilby:
A somewhat more reasonable way to generate the error codes

Generate using python instead of perl
Error codes in executionerrors.h and parseerrors.h are now of the form
errorCode // "errorString"
where errorCode is a unique identifier (enum)
and errorString is the human-readable error message in quotes

This makes the addition of new error codes less error-prone
and is a start to getting rid of perl in the build process.


bwmilby:
Not needed.  A squash merge on the original PR will be fine.

PR #393, opened 2026-07-18
python as linux alternate language

mwieder:
This patch allows the use of python(n) as an alternate language on linux.
It implements "the alternatelanguages" and "do xyzzy as python".

Example:
on mouseUp
   local tScript
   put "import sys" & cr into tScript
   put "print('Hello from temporary Python script')" & cr after tScript
   put "print('argv:', sys.argv)" & cr after tScript
   
   do tScript as "python3"
   put the result into field 1
end mouseUp

### DIFF
--- a/engine/src/dsklnx.cpp
+++ b/engine/src/dsklnx.cpp
@@ -2164,16 +2164,224 @@ virtual real64_t GetCurrentMicroseconds(void)
 #endif
     }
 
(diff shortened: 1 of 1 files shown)


emily-elizabeth (approved):

emily-elizabeth:
Tested on macOS 26.5.2, Windows 11 and Ubuntu 25.04
Obviously the code only ran on Ubuntu.

PR #394, opened 2026-07-18
392 toolbar missing in script editor

### LINKED ISSUE #392: Toolbar missing in Script Editor
Commit e91574855cfdd71f261f391d0462dad4648077b2 makes a change that results in the Script Editor no longer having a toolbar.  The amount of space used by the toolbar is blank (white in dark mode) at the **bottom** of the window.  When the window is resized, the bottom does get painted and used but the top remains the same with no toolbar.  The button to show/hide the toolbar only moves things up/down by 1 or 2 pixels.

As a test, I checked out `main` and did a revert of the named commit and the toolbar was now available.  The button to show/hide the toolbar still doesn't really work correctly though.  The toolbar does disappear, but the content of the window does not adjust/move up.

emily-elizabeth:
Background
engine/src/osxstack.cpp implements the macOS-specific MCStack window operations — applyscroll(), clearscroll(), setgeom(), platform_openwindow(), and related functions. The file was guarded out on arm64 with #if !defined(__arm64__) && !defined(__aarch64__), and engine/src/mac-stubs-arm64.mm provided no-op replacements. This meant applyscroll() — the mechanism that shrinks a stack's visible rect to hide a menubar group behind the native macOS menu bar — was silently inactive on Apple Silicon.
Changes
engine/src/osxstack.cpp
Removed the arm64 exclusion guard. The file uses Window as typedef MCPlatformWindowRef (a plain C++ pointer), not the Carbon WindowRef, so there are no arm64-incompatible types. osxprefix.h already provides HITheme stubs for arm64 builds.
engine/src/mac-stubs-arm64.mm
Removed the // ── Stack window ops block (~60 lines of duplicate no-op MCStack:: implementations) that are now superseded by the real osxstack.cpp code.
engine/src/ (8 files)
Removed leftover fopen/fprintf debug trace blocks that wrote to hyperxtalk-arm64-startup.log across desktop.cpp, dskmain.cpp, exec-engine.cpp, external.cpp, mac-core.mm, mode_development.cpp, stack.cpp, and stack2.cpp.
ide/Toolset/palettes/script editor/behaviors/revsestackbehavior.livecodescript
Fixed a toolbar positioning regression caused by applyscroll() now being active. The existing HXT fix in actionResizeStack placed the Toolbar group at card y=0, which was correct when applyscroll() was a no-op (no rendering shift). With applyscroll() active, the engine's rendering transform shifts all content up by m_scroll pixels, so card y=0 maps to a point above the visible window. The fix positions the Toolbar at tCardTop + the vScroll of me instead, which equals m_scroll and places the Toolbar at the first visible pixel of the window. This is consistent with the existing seResizeCardBottom fix, which already adds the vScroll of me to tCardBottom for the same reason.
Testing
Open the Script Editor on an Apple Silicon Mac and verify the Toolbar (Compile / Undo / Redo / Run / Stop / Step buttons) appears at the top of the window with no blank strip above or below.


bwmilby:
Verified this corrects the toolbar visibility.  This also addresses the show/hide toolbar issue.
When in dark mode, there is still a band at the bottom of the window that does not get the dark background.  As soon as the window is resized the area is painted correctly.

emily-elizabeth:
You mean like this? I'd recommend opening a new ticket

image

PR #396, opened 2026-07-20
Added javascript to the list of linux alternate languages

mwieder:
Added javascript to the list of linux alternate languages if node.js is found in the path.
Tasks can now be delegated to python and javascript enabling a wider range of external libraries.
* Fixed a possible commandline buffer overflow.

### DIFF
--- a/engine/src/dsklnx.cpp
+++ b/engine/src/dsklnx.cpp
@@ -2165,7 +2165,6 @@ virtual real64_t GetCurrentMicroseconds(void)
     }
 
 // return 1 for true, 0 for false, -1 on error
-//TODO: fix the occasional buffer overflow
 int is_language_installed(const char *prog) {
     int pipefd[2];
     if (-1 == pipe(pipefd)) {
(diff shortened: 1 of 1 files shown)


emily-elizabeth:
Could the JavaScript functionality be brought to the other two platforms?

emily-elizabeth (approved):

mwieder:
I suppose. The grunt work there is in the is_language_installed function. It's possible that it could work as is on osx or with a bit of tweaking. I would expect a totally different approach on Windows. And maybe Windows would need a different temp directory in MCS_alternate_shell. But node.js is available on all the desktop platforms.
One other thing: I'm assuming that python(x) is installed by default. Don't know if that's true on Windows.

Updated: Seems Windows does not have python or node or perl installed by default.
And OSX used to have python by default but no longer does. OSX currently has perl but may not for long.
None of that prevents users from homebrew-installing extra language packages or whatever it takes on Windows.

Yet another update: Windows doesn't have mkstemp. There's an ugly replacement for it here
[https://stackoverflow.com/questions/6036227/mkstemp-implementation-for-win32](url)
...and it seems the mac build supports javascript as an alternate language already.
(dskmac.cpp:getosacomponents)
I looked at dskw32.cpp and now my brain hurts.

mwieder:
I added ruby, lua, and go as alternate languages.
Also made it a bit more modular so may pieces could be cannibalized for other platforms as needed.

PR #400, opened 2026-07-22
[[Issue 343]] Clean JSON export string

### LINKED ISSUE #343: jsonExport emits invalid JSON
jsonExport does not clean data and allows null characters in the output (this issue originates from code in the original repository).  It should emit proper JSON or throw an error.

jsonImport does not handle the invalid null bytes in the data and will throw an error.  It does not handle it gracefully (does not report that the data is invalid).

Adding the following above the unsafe call will clean the data before passing over to the binding code.  This was just done to demonstrate that it was an issue with the input data and is not something that should be added.  Just fixing the null would still result in errors so the entire control space is cleaned:
```
  variable tChar as String
  repeat for each char tChar in pJson
    if the code of codeunit 1 of tChar <= 0x1f then
      replace tChar with "/u" & (the code of tChar converted to base 16) in pJson
    end if
  end repeat
```

Technically since the input file is invalid with those characters, it would be better to throw an error.  The error should be graceful and let the user know the data is not valid.  The output side needs to be corrected.  It should either encode the data or report an error.

A pretty print option would be nice.  Raw JSON isn't going to work well with GitHub or other version control systems and is not easily read by humans.

Issue discovered by exporting/importing my ToutureTest stack as a HyperXTalk Project.

bwmilby:
Updated JSON export to ensure that non-printable characters are not exported without conversion.

Rename script file to match stack name (jsonexport)
Add `_cleaningNonPrintableChars` from PhotonJSON
Add `_convertAsciiToUnicodeNum ` from PhotonJSON

Normalize "JSON" in variable names and handler names.  Standard for CamelCase is to only capitalize the first letter of acronyms so `tJSON` becomes `tJson`.

Since the cleaning of non-printable characters happens in the script library, moved `JsonExportPretty` handler to the script library to ensure consistent results.

Closes #343


emily-elizabeth:
We can use yyJSON to do pretty print as well. We currently use it for importing JSON

bwmilby:
We already implemented it in LCB.  All this does is put the guard in place that cleans the data.  Due to where the clean takes place, I moved the duplicate handler (it is essentially an alias - almost).

yyJSON uses null delimited strings which means that data has to be checked before passing.  This issue was discovered when importing data exported by this library.  Library failed without any error importing.

emily-elizabeth:
Sounds good.

emily-elizabeth (approved):

bwmilby:
I'm not sure how much you looked at using yyJSON for the export, but I'm thinking that due to how the library is used it won't save too much.  We would still need to iterate in LCB and build the doc a piece at a time.  Once the doc is built, then we could have the library emit the JSON text, with or without pretty printing it.  There isn't a built in way to reformat JSON (would have to parse JSON to a doc and then export, but that could probably be done all in c).  Since we can't easily clean the data in bulk, we would have to do a clean on each text element before passing to the library (minimum would be to fix the nulls in strings).  I think that a longer term goal should be to put the JSON formatting into the engine so it can have access to the native MC data types.  We'd also need to implement "is strictly" for array and list so that we can do two way transport of JSON with fidelity since `["a", "b"]` is different from `{"1":"a","2":"b"}` but the LC/HXT arrays will look the same.  Type casting would also be needed (e.g. `local tList as a list`).

PR #401, opened 2026-07-23
Update revsblibrary.livecodescript

### LINKED ISSUE #364: Attempting to use custom icons fails on Mac OS ARM
The icons section of the build settings allows one of three options to be used.  One of these is "Choose" which allows a .icns file to be selected.  Unfortunately when using this option the full path to the .iicns file is not entered into the path field and the subsequent build fails to find the icon file so reverts to the default icon.  The field does not allow the text to be entered or edited.

Image

An unimportant further issue is that the settings pane still displays Livecode branding.

Image

emily-elizabeth:
Problem

When a custom .icns file was selected in the macOS section of Standalone Application Settings, it was silently discarded on every build. The icon would revert to the default GenericApplication.icns regardless of the user's choice.

Root Cause

In ide-support/revsblibrary.livecodescript, the revSBDefaultStandaloneSettings handler unconditionally overwrote OSX,iconFile with the generic icon path every time it was called — which happens on every build via revSBGetSettings. Unlike every other default setting in the same handler, it had no if xSettingsA["OSX,iconFile"] = "" then guard.

Fix

Added the missing guard so the generic icon is only set as a default when no custom icon has been configured:

livecode
if xSettingsA["OSX,iconFile"] = "" then put tName & "/GenericApplication.icns" into xSettingsA["OSX,iconFile"]

Testing

Set a custom .icns file in Standalone Settings → macOS section
Build a macOS standalone
Verify the app bundle uses the custom icon instead of the default

### DIFF
--- a/ide-support/revsblibrary.livecodescript
+++ b/ide-support/revsblibrary.livecodescript
@@ -876,7 +876,7 @@ on revSBDefaultStandaloneSettings pStack, @xSettingsA
    end if
    // HXT
    put revEnvironmentRuntimePath() & "/Mac OS X/arm64/Support/Sample Icons" into tName
-   put tName & "/GenericApplication.icns" into xSettingsA["OSX,iconFile"]
+   if xSettingsA["OSX,iconFile"] = "" then put tName & "/GenericApplication.icns" into xSettingsA["OSX,iconFile"]
    
    if xSettingsA["answerDialog"] = "" then put kAnswerDialog into xSettingsA["answerDialog"]
    if xSettingsA["cursors"] = "" then put kCursors into xSettingsA["cursors"]


emily-elizabeth:
This is good for testing

bwmilby:
The code in this change is good to go.

There are some things in the `makefile` for `package-mac-bin` that are no longer needed.  It seems that the icon file was moved at some point so lines 515-519 are not needed.  The copy in line 514 creates the path that 516 created.  The file referenced in 517 does not exist at that location, but was already copied in 514.

bwmilby (approved):

emily-elizabeth:
Lines 515-519 are removed.

emily-elizabeth:
I forgot to push the changes on this branch, so I did it on `main`

PR #402, opened 2026-07-23
[[Issue 107]] Remove `revdatabaselibrary`

### LINKED ISSUE #107: Rip out db library OR start actually using it
the dblibrary gets installed with every built app whether you want it or not (this is a LC fossil). It inserts itself into the message path as a frontscript.

the original idea was to bind objects automagically to databases. LC eventually got rid of the binding, but the stack still gets forcibly installed in every built project.

we should either not install it no matter what or (and?) bring back binding.

bwmilby:
Delete `revdatabaselibrary.livecodescript` file.
Remove reference in `home`
Replace loading the library with setting db driver path in `revidelibrary`

Did some research on the library and none of the handlers are ever used in any of the IDE code.  From the comments in the code it seemed like groups were going to be used to store the information for database connections.  If someone really needs it, we could just package it as an optional library that could be loaded by hxtlib.

Closes #107


emily-elizabeth (approved):

emily-elizabeth:
Compiled and ran on macOS 26.5.2, Windows 11 and Ubuntu 25.04

PR #407, opened 2026-07-27
Folding 2

BerndN:


BerndN:
This should add code folding to HyperXTalk, tested by temporarily installing into HXT 0.9.15

emily-elizabeth:
I'm seeing this instead of the arrows. The folding still works, just a visual glitch

image

BerndN:
I have no idea where that comes from.
Have a look at https://github.com/emily-elizabeth/HyperXTalk/pull/407/commits/ea0788450e517f0fc7503608cbee7d0bc4993541

emily-elizabeth:
Maybe a font issue?

bwmilby:
I think that scripts pretty much should be ASCII.  Even when changing those 2 chars to use `numtocodepoint` I still get the wrong indicator.  What OS was this developed on?

BerndN:
> I think that scripts pretty much should be ASCII. Even when changing those 2 chars to use `numtocodepoint` I still get the wrong indicator. What OS was this developed on?

It was developed on MacOs Sequoia 15.7.7. The folding indicators show up correctly in a preview app that Emily has kindly provided. The font I use is source code pro. I am at a loss.
In GitHub (see reference above) they are displayed correctly.

bwmilby:
The symbols that Emily and I are seeing are the same as what you get when encoding those characters as UTF8.

bwmilby:
They show up fine in VS Code for me as well.  In the IDE I see the encoded values in the script.  What baffles me is that I get the same thing using `numtocodepoint` (which I would have thought to be the solution).  If I change them to ASCII, they display whatever I changed them to.

bwmilby:
Making this change everywhere showed the proper characters in the SE gutter:
```
   put textdecode(sePrefGet("editor,foldingopen"),"UTF8") into tFoldingOpen
   put textdecode(sePrefGet("editor,foldingClosed"),"UTF8") into tFoldingClosed
```
It worked with and without making the `numtocodepoint` change as well.  I think that we should do both (as long as it works everywhere) to avoid putting unicode chars in the actual code file.
This shows that the problem is related to unicode encoding/decoding.  I'm still not certain if this is the best solution.

emily-elizabeth:
Without the changes from Brian, this is what it looks like on Windows

image

BerndN:
> Making this change everywhere showed the proper characters in the SE gutter:
> 
> ```
>    put textdecode(sePrefGet("editor,foldingopen"),"UTF8") into tFoldingOpen
>    put textdecode(sePrefGet("editor,foldingClosed"),"UTF8") into tFoldingClosed
> ```
> 
> It worked with and without making the `numtocodepoint` change as well. I think that we should do both (as long as it works everywhere) to avoid putting unicode chars in the actual code file. This shows that the problem is related to unicode encoding/decoding. I'm still not certain if this is the best solution.

Brian, if you mean by actual code file the script then tFoldingOpen and tFoldingClosed are only used in the field "Folding" right next to field "Numbers" in the gutter. It does not touch the actual code in field "Script"

I changed 2 lines in revseutilities:

From 

   put "▽" into tPrefs["editor,foldingOpen"]
   put "▶" into tPrefs["editor,foldingClosed"]

To
   put textDecode("‚ñΩ","UTF8") into tPrefs["editor,foldingOpen"]
   put textDecode("‚ñ∂","UTF8") into tPrefs["editor,foldingClosed"]

and it is working for me in field "Folding" of the gutter.

Would that be a solution?

bwmilby:
It does not correct the issue on my seat.  I still end up seeing the encoded values in the gutter.

I changed all 5 spots where the 2 prefs are read and put into those variables in the editor behavior.

bwmilby:
Here's a potentially better solution (`revseutilities`) since it touches less code:
```
function sePrefGet pTag
   if pTag is "editor,font" and \
         sPrefs[pTag] is empty and \
         "Source Code Pro" is among the lines of the fontNames then
      return "Source Code Pro"
   end if
   if pTag is "editor,foldingopen" or pTag is "editor,foldingClosed" then
      return textdecode(sPrefs[pTag],"UTF8")
   end if
   return sPrefs[pTag]
end sePrefGet
```

@BerndN 
What happens if you just make this change:
```
   put numtocodepoint(9661) into tPrefs["editor,foldingOpen"]
   put numtocodepoint(9654) into tPrefs["editor,foldingClosed"]
```
My guess is that you don't see any change and it still works like it did before.  For me, I see the same thing whether or not I make that change and I need the above change (sePrefGet) to see the correct fold indicators in the gutter.

BerndN:
>   put numtocodepoint(9661) into tPrefs["editor,foldingOpen"]
>    put numtocodepoint(9654) into tPrefs["editor,foldingClosed"]

If I make that change to handler "sePrefGetDefaults" AND NOT make the changes to handler "sePrefGet" then I get the expected result.

If I do the suggested changes to handler "sePrefGet" I get all question marks instead of the folding indicators.

What operating system to you use?

bwmilby:
I'm on Tahoe 26.5.2

emily-elizabeth:
I'm on macOS 26.5.2 (Tahoe) and Windows 11. I haven't tested any of the new code yet

bwmilby:
If I add this to `seGetPref` (before the return):
```
   if pTag is "editor,foldingopen" then
      put sPrefs[pTag] && numtocodepoint(9661) && (sPrefs[pTag] is numtocodepoint(9661))
   end if
```
I get `‚ñΩ ▽ false` and I'm guessing that you will get `▽ ▽ true`
The other difference is that I'm using 0.9.16 (from main).  All changes made to scripts on disk and then I launch IDE to test.

bwmilby:
I don't like it, but this does work for me:
```
   if (pTag is "editor,foldingopen" or pTag is "editor,foldingClosed") and \
         len(sPrefs[pTag]) = 3 then
      return textdecode(sPrefs[pTag],"UTF8")
   end if
```
Here is the debug that I put before it:
```
   if pTag is "editor,foldingopen" then
      put len(sPrefs[pTag]) && sPrefs[pTag] && numtocodepoint(9661) && (sPrefs[pTag] is numtocodepoint(9661))
   end if
```
For me, I get `3 ‚ñΩ ▽ false` and I'd imagine you would see `1 ▽ ▽ true`.

BerndN:
> I get ‚ñΩ ▽ false and I'm guessing that you will get ▽ ▽ true

I get as you guessed
▽ ▽ true

So what can we do remedy the situation. It makes my head spin. Especially that @emily-elizabeth also sees "funny" things on Tahoe without importing. 

Since it works for me since ages and many people have used "code folding" as a patch and nobody mentioned this artifact before I am confused. 
What would be the difference between MacOS Tahoe and Sequoia in this regard?

bwmilby:
We have a couple workarounds.  If the symbol length is 3 then it needs be decoded before being returned, otherwise the value can just be returned.

We could also set the defaults this way:
```
   put textencode(numtocodepoint(9661),"UTF8") into tPrefs["editor,foldingOpen"]
   put textencode(numtocodepoint(9654),"UTF8") into tPrefs["editor,foldingClosed"]
```
This would make the length check unnecessary since we will always need to decode.

What I don't understand is why, on my machine and the Windows machine that Emily checked, the unicode char is being stored in the array encoded.

emily-elizabeth:
So what is the code to use?

bwmilby:
I'm doing a build on Linux to check.  I can push an update to the PR with my suggestion.

emily-elizabeth:
Sounds good. I will wait for it so I can test on Windows.

bwmilby:
Tested my push in Tahoe and Linux.  Verified Linux had same issue prior to push.

emily-elizabeth:
Will test this out today

emily-elizabeth:
This is what I'm seeing on Windows 11, so good work Brian 😄 

image

BerndN:
It does not work on my Sequoia Mac, just question marks instead of Open/Close.

emily-elizabeth:
Perhaps we need to do a conditional based on the `systemVersion`?

BerndN:
My gut feeling is that there is something else going on, but I have no idea what. It just sounds funny that something as basic as this should all of a sudden work differently between two consecutive versions of Mac OS.

emily-elizabeth:
Well, macOS 26, Windows 11 and Linux all show the same thing, so technically it's macOS 15 that's out of sorts 😉  I think it's a really useful feature and would like to include it for 0.9.16 so if we need a conditional check to get that done then we should use it.

```
set the itemDelimiter to "."
if (the platform = "macOS") AND (item 1 of the systemVersion < 26) then
-- do it Bernd's way
else
-- do it Brian's way
end if
set the itemDelimiter to COMMA
```

bwmilby:
@BerndN are you making the changes on disk and then launching the IDE?

bwmilby:
> It does not work on my Sequoia Mac, just question marks instead of Open/Close.

That is with both changes (default and get)?  If you add the debug put before the return, what Len are you seeing?  To get ? it is likely 1 which would mean that even when explicitly encoding it is not being stored encoded.

BerndN:
I get len = 1 with both patches: to sePrefGet and sePrefGetDefaults

BerndN:
What happens if you make a new stack and 1 button and set the script to 

on mouseUp
   local tArray
   put "▽" into tArray[1]
   put tArray[1]
end mouseUp

?

emily-elizabeth:
I get the down arrow

image

BerndN:
I deduce from this that it is specific to the inside of IDE. Not HXT unicode handling in general?

emily-elizabeth:
We haven't changed anything in regards to Unicode.

bwmilby:
[UTF-stuff.livecode.zip](https://github.com/user-attachments/files/30508828/UTF-stuff.livecode.zip)
Can you try this stack out?  I don't understand why the code isn't working on Sequoia with explicit encode/decode.

emily-elizabeth:
This is what I get on macOS 26.5.2

image

bwmilby:
> I get len = 1 with both patches: to sePrefGet and sePrefGetDefaults

That is incorrect.  It should be 3.  The encoded UTF of that symbol takes 3 bytes.  I guess a better spot to put the debug would be immediately after setting the value in the defaults handler.

The workaround is actually simpler than testing for OS:
```
   if (pTag is "editor,foldingopen" or pTag is "editor,foldingClosed") and \
         len(sPrefs[pTag]) = 3 then
      return textdecode(sPrefs[pTag],"UTF8")
   end if
```

I was really hoping to get around checking the length by ensuring the stored value is encoded.

bwmilby:
> This is what I get on macOS 26.5.2
> 
I get the same on Intel Sequoia using LC 9.6.12 and LC 10.0.2
In Linux I get something slightly different but expected.  The Mac encoded values are not valid so I get ?? for "decode,foldingOpen".

BerndN:
This is what I get;

Opensymbol

bwmilby:
Now I'm really confused.  Based on your results, I don't see why you would get `??` in the SE.
```
on mouseup
   local tPrefs
   put textencode(numtocodepoint(9661),"UTF8") into tPrefs["fo"]
   put len(tPrefs["fo"]) && tPrefs["fo"]
end mouseup
```
For the above, I get: `3 ‚ñΩ`
This is essentially what my stack was checking (last line in the field).  My stack also included your test as well.

bwmilby:
I think I figured it out.  It is the preferences stack holding a previous value causing the inconsistency.

bwmilby:
Try this... Open the IDE and before opening the SE, open the message box and do these 2 commands:
```
revIDESetPreference "cScriptEditor,editor,foldingOpen", empty
revIDESetPreference "cScriptEditor,editor,foldingClosed", empty
```
Then open the SE.  I went back to almost the original code (still used numtocodepoint).  At this point either version should work correctly.  I think leaving UTF8 characters out of the script is safest (fine to put them in comments knowing that if the BOM is removed they won't render properly in the IDE).  Tested on Linux as well.

What I think happened is that with the BOM removed, when we opened the file to test initially, it saved the encoded value because the script was not opened as UTF8 (text editors guess, but the IDE relies on the BOM - one of my first PRs :) ).  Once we had that value saved in the prefs, we were always seeing a 3 byte encoded value vice the UTF8 single code point character.

BerndN:
I deleted my prefs and it now works as expected. Still using

in sePrefsgetDefault
   put textencode(numtocodepoint(9661),"UTF8") into tPrefs["editor,foldingOpen"]
   put textencode(numtocodepoint(9654),"UTF8") into tPrefs["editor,foldingClosed"]

in sePrefGet
   if (pTag is "editor,foldingopen" or pTag is "editor,foldingClosed") then
      put len(textdecode(sPrefs[pTag],"UTF8"))
      return textdecode(sPrefs[pTag],"UTF8")
   end if

BerndN:
Now I changed to your new revert PR

in sePrefsgetDefault
   put numtocodepoint(9661) into tPrefs["editor,foldingOpen"]   -- "▽"
   put numtocodepoint(9654) into tPrefs["editor,foldingClosed"] -- "▶"

in sePrefGet
  removed special handling of folding

after resetting the prefs it worked flawlessly 

revIDESetPreference "cScriptEditor,editor,foldingOpen", empty
revIDESetPreference "cScriptEditor,editor,foldingClosed", empty

What happens when you set the BOM again? Would this whole dance be unnecessary? 

As it is now I think I can live with it quite well because it works without branching according to current system.

I would like to say a huge THANK YOU to you. Who would have "thunk" :)

bwmilby:
The BOM (byte order mark) describes the data in the file.  VS Code assumes UTF8 if not otherwise specified (At least for this file it did).  The IDE does not, so the file is loaded as ASCII which means UTF data is displayed with the encoded values.  (Originally LC couldn't handle a BOM on `livecodescript` files but now it can.)  Due to this situation, when we looked at the file in VS Code it looked fine, but was a bit corrupted in the IDE.  Setting the BOM initially would have avoided all of the issue since the pref would have never been saved as encoded values.  Since the file initially loaded in a way to save encoded values, I was chasing my tail to address that issue and missed the root cause.

Your system's refusal to work caused me to look deeper, so I'm glad it didn't work.

Using `numtocodepoint` is insurance against the BOM getting removed accidentally in the future.  With a BOM, using UTF8 literals is possible.

I'm just happy that we were able to figure it out.

emily-elizabeth (approved):

emily-elizabeth:
After deleting my preferences files I am seeing the arrows on macOS 26.5.2 and Windows 11.

emily-elizabeth:
Is this good to merge now?

BerndN:
I think all issues are resolved and for me it is good to merge. Everything works as expected.

I guess Brian should also agree which he implicitly did. 
Since we three were the only ones to haver their prefs filled with wrong prefs all other should have no issues after Brians pull requests are merged.

PR #414, opened 2026-08-02
376 backdrop across multiple monitors altab

emily-elizabeth:
Fix ALT+TAB blocked by keep_above when backdrop is active

**Problem**
When the HXT backdrop was active, pressing ALT+TAB would show the window switcher and allow selecting another application, but that application would never come to the front. The backdrop and stack windows had _NET_WM_STATE_ABOVE set, placing them in the WM's ABOVE layer. Mutter (and Muffin) physically cannot raise a NORMAL-layer window above an ABOVE-layer window, so the selected application remained buried.

**Solution**
Remove keep_above entirely from all backdrop and stack windows. Everything stays in the NORMAL WM layer so the WM can freely raise other applications.

HXT windows are re-raised via two paths:

_NET_ACTIVE_WINDOW PropertyNotify filter — a GDK event filter on the X11 root window fires whenever the active window changes. If the newly active window belongs to HXT, backdrops and stacks are re-raised. If it belongs to another application, they are left alone.
Backdrop button-press — clicking the backdrop doesn't change _NET_ACTIVE_WINDOW (because accept_focus=FALSE), so the existing button-press handler calls reraise_stacks_above_backdrop() → backdrop_focus_gained() to correct Z-order directly.

**Key details**
The _NET_ACTIVE_WINDOW filter ignores xwin=0, which is a transient state the WM emits during ALT+TAB window selection. Reacting to it disrupted the WM's own stacking.
On Muffin (Cinnamon), _NET_WM_WINDOW_TYPE_NORMAL is explicitly set on backdrop windows to prevent Muffin inferring a higher window layer from Motif hints.
On Mutter, stacks are transient_for the primary backdrop, but extra per-monitor backdrop windows are separate. All stacks are raised explicitly in backdrop_focus_gained() to guarantee correct Z-order across all monitors on all WMs.
The Muffin stacking timer (hxt_check_stacking, 250 ms) now runs whenever Muffin is detected, not only when extra backdrops are present.

**Files changed**
engine/src/lnxdcs.cpp — removed keep_above, added _NET_ACTIVE_WINDOW filter, updated backdrop_focus_gained, backdrop_focus_lost, assignbackdrop, enablebackdrop
engine/src/lnxdclnx.cpp — removed redundant reraise_stacks_above_backdrop call from GDK_FOCUS_CHANGE handler (now handled by the root window filter)
engine/src/lnxdc.h — added declarations for reraise_stacks_above_backdrop, backdrop_focus_gained, backdrop_focus_lost

**Tested on**
Ubuntu (Mutter), Linux Mint (Muffin), Debian (Mutter), Fedora (Mutter), Debian MATE (Marco)


mwieder (approved):
I don't like or use the backdrop 'feature' so this seems like a lot of bloat to carry around.
But it works as advertised on my three monitors, so I'm good with it.

PR #418, opened 2026-08-03
[mac] Fix extra backdrop windows covering HXT windows on external displays

emily-elizabeth:
Extra backdrop windows were set to kCGNormalWindowLevel - 1, placing them permanently below all normal-level windows including those from other apps. This meant foreign windows on the external display always appeared above the backdrop even when HXT was the active application.

Switch extra backdrop windows to NSNormalWindowLevel and use per-display smart ordering: find the back-most HXT window on each display and slot the extra backdrop just below it. If no HXT window is on that display, orderFrontRegardless covers other apps without activating any foreign application.

Also exclude extra backdrop windows from the main ordering loop since they are now at NSNormalWindowLevel and would otherwise interfere with primary display z-ordering.

### DIFF
--- a/engine/src/mac-core.mm
+++ b/engine/src/mac-core.mm
@@ -1580,11 +1580,21 @@ void MCMacPlatformSyncBackdrop(void)
     [CATransaction begin];
     [CATransaction setDisableActions: YES];
 
+    // Build a set of extra backdrop NSWindow handles for quick lookup below.
+    NSMutableSet *t_extra_set = [NSMutableSet set];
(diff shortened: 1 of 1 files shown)


bwmilby (approved):
If I have windows interleaved (stack, foreign app, stack) and turn on backdrop, then the backdrop is behind the furthest back HXT window and the other app remains visible.  If I click on the HXT icon in the dock, then only HXT windows will be visible.  Same situation on the primary display.  This does seem to match how Mac apps work though - clicking on a window for an app won't bring them all to the front (unless "group windows by application" is turned on).  May be worth a mention in the docs on the feature.

PR #420, opened 2026-08-03
Fix autocomplete popup appearing on wrong screen and offset vertically

### LINKED ISSUE #415: autocomplete popup appears on the wrong screen
> I find the autocomplete popup menu annoying, especially since it pops up on a different monitor from my script editor

emily-elizabeth:
Two fixes in ideautocompletelibrary.livecodescript:

1. Wrong screen: 'the screenRect' always returns the primary display rect. When the script editor is on an external monitor the popup position was clamped to the primary screen bounds, causing it to appear on the wrong display. Fix: iterate 'the screenRects' to find the display containing the caret and use that rect for boundary clamping.

2. Vertical offset: the toolbar fix in #394 positioned the Toolbar at tCardTop + vScroll to compensate for applyscroll() shifting card content up by vScroll pixels on arm64. The caret Y calculation used windowY + cardY without accounting for this offset, placing the popup vScroll pixels too low. Fix: subtract the vScroll of the enclosing stack from tWinY on macOS so card-space coordinates convert correctly to screen coordinates.

### DIFF
--- a/ide/Toolset/libraries/ideautocompletelibrary.livecodescript
+++ b/ide/Toolset/libraries/ideautocompletelibrary.livecodescript
@@ -346,9 +346,18 @@ private command ideAutocomplete_ShowPopup pEditorId, pMatches
    local tX, tY
    ideAutocomplete_GetCaretScreenPos pEditorId, tX, tY
 
-   -- Keep the popup on screen
+   -- Keep the popup on the same screen as the caret.
+   -- 'the screenRect' always returns the primary display rect; use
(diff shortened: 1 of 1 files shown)


emily-elizabeth:
tested on macOS 26.5.2 and Windows 11

emily-elizabeth:
Linux is on its own branch due to it being Linux and giving me nothing but headaches

bwmilby (approved):
Looks good.  Tested on Mac 26.5.2

PR #421, opened 2026-08-03
415 lin autocomplete popup appears on the wrong screen

emily-elizabeth:
Linux: fix autocomplete popup appearing behind script editor

The autocomplete popup (ideAutoCompletePopup, style=palette) was appearing behind the script editor when the backdrop was active.

Root cause

The ALT+TAB fix (PR #414) moved backdrop windows from the X11 ABOVE layer to the NORMAL layer. However, assignbackdrop() was still calling gdk_window_set_transient_for(popup, backdrop) for all registered windows including WM_PALETTE. On Mutter, a transient child inherits its parent's stacking layer — so the popup was being pulled from ABOVE (where openwindow()'s pre-map _NET_WM_STATE_ABOVE had placed it) down into NORMAL, ending up behind script editors that legitimately remain in ABOVE.

Fix

Skip gdk_window_set_transient_for for WM_PALETTE windows in assignbackdrop(). Palette windows carry their own _NET_WM_STATE_ABOVE and should float above all stacks independently of the backdrop's layer membership.

WM compatibility

Mutter (GNOME/Ubuntu): palette retains its own _NET_WM_STATE_ABOVE, floats correctly above stacks
Muffin (Cinnamon): already skipped transient_for via the s_wm_is_muffin branch — no change in behaviour
Marco (MATE): same as Muffin path — no change in behaviour

Files changed

engine/src/lnxdcs.cpp — assignbackdrop(): add p_mode != WM_PALETTE guard on transient_for call

### DIFF
--- a/engine/src/lnxdcs.cpp
+++ b/engine/src/lnxdcs.cpp
@@ -321,6 +321,7 @@ void MCScreenDC::reraise_stacks_above_backdrop()
 
 static x11::Atom  s_net_active_window_atom = None;
 static bool       s_active_win_filter_installed = false;
+static x11::Window s_last_active_xwin = None;
 
 static GdkFilterReturn hxt_active_win_filter(GdkXEvent *p_xevent,
                                               GdkEvent  * /*p_event*/,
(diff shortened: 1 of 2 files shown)


emily-elizabeth:
Tested on Ubuntu 25.04, Linux Mint, Debian 13.5 (GNOME), Fedora, Debian 13.5 (MATE)

mwieder (commented):
Looks good.

PR #425, opened 2026-08-05
Fix pop without push causing silent SIGSEGV crash

### LINKED ISSUE #408: Pop card crashes 9.15
> pop card crashes 9.15 is there is not a prior push card

emily-elizabeth:
- popcard() now returns nullptr when push stack is empty instead of dereferencing a potentially invalid pointer
- Added null checks in MCInterfaceExecPop and MCInterfaceExecPopToLast; both now throw EE_POP_EMPTYSTACK on empty stack
- Added EE_POP_EMPTYSTACK error (repurposed UNUSED slot 396) with message "pop: push stack is empty"


emily-elizabeth:
Now throws an error

image

bwmilby:
Very interesting.  Looking at the code, it seems the original intent was to just return the current card if the push stack was empty.

bwmilby (approved):
Compiled on Mac, test success.

bwmilby:
NOTE:  Every test in the test suite always does a `push card` at some point before doing a `pop card`.

emily-elizabeth:
This is what the documentation states

> If the list is empty when you use the pop command, HyperXTalk goes to the Home card. If you are running in the HyperXTalk development environment, this is the first card of the "license.rev" stack, which is the splash screen that appears when you start up HyperXTalk. (Click this card to close the stack.) If you are running in a standalone application, the Home card is the first card of the standalone's main stack.

bwmilby:
Pondering this a bit, I wonder if it is due to the IDE Home now being a script only stack.  If never visited/opened, the `card - cards - currcard` values could all still be NULL.  Since there was no check for NULL, it would make sense to cause a crash.

PR #426, opened 2026-08-05
Fix 'the properties of toolbar' crashing on item properties

### LINKED ISSUE #388: `the properties` of toolbar crash
`get the properties of toolbar "myToolbar"` will crash HyperXTalk

Looking at `toolbarprops`, the following properties generate an error when attempting to get:
- itemEnabled
- itemIcon
- itemLabel
- itemStyle
- itemTooltip

This is what is causing the crash. `the properties` only works for individual properties (generally ones that can be rendered as a string).  When an invalid property is included in the PropList constant, it will just crash when accessed (discovered while working on Widget PropList).
```
HyperXTalk/mac-bin/HyperXTalk.app/Contents/MacOS/HyperXTalk exiting on signal 11
```

One possible solution would be to add accessors to get each of those 5 properties back as a list the same way that is used for `GetItemNames` (add an `s` to the end of each for this version which would be what is included in the `toolbarprops`).  The downside here would be that setting the properties this way would not be supported unless that direction was also implemented.  Supporting setting this way would mean that the tooltip could not be multiple lines.

Also, too much debug output is generated from Toolbar use written to stderr.  The debugging probably should be limited to when compiling for debug purposes.

emily-elizabeth:
The 5 per-item array properties (itemEnabled, itemIcon, itemLabel, itemStyle, itemTooltip) require an item-name index and cannot be serialised as a scalar string. Including them in toolbarprops caused a SIGSEGV when 'the properties' iterated the PropList.

Fix: add 5 read-only list getters that return all items' values as a newline-delimited string (matching the itemNames pattern), and replace the 5 singular entries in toolbarprops[] with these:

  itemEnableds  -- "true"/"false" per item
  itemIcons     -- icon name per item
  itemLabels    -- label per item
  itemStyles    -- style per item
  itemTooltips  -- tooltip per item

New properties are also exposed in the lexer so scripts can read them directly (e.g. get the itemLabels of toolbar "myToolbar").


emily-elizabeth:
```
on MouseUp
   get the properties of toolbar "mainBar"
   get the itemIcons of toolbar "mainBar"
end MouseUp
```

bwmilby (approved):
Compiled/tested on Mac
With this test, I am able to remove my guard on the toolbar object and can now use `the properties`.

NOTE:  There is still quite a bit of debugging emitted to the console when using toolbars (mostly starts with `[MCToolbar]` but there are other items)

emily-elizabeth:
I'll take a look at the debugging code and remove it

emily-elizabeth:
Debug logging has been removed.

PR #427, opened 2026-08-05
Fix painted images disappearing when navigating away from card

### LINKED ISSUE #383: painted images disappear when closing the card/stack
Not that they work: if you paint on the card, it doesn't use the "suitable paintimage" fix, so when you fully close and reopen the stack, the painted image is gone (ceases to exist on the stack). Happens when navigating away from the card, so no need to close the stack

Image

Image

emily-elizabeth:
Root cause: MCCard::createimage() never called setid() on the new image. After card close, MCObjptr::clearref() nulls the cached pointer; on reopen getref() re-binds by ID — with ID=0 the lookup always fails and the image is permanently lost.

Fix: assign a proper stack-unique ID in createimage() before the image is registered via newcontrol(), matching what every other tool's creation path already does.

Also call finishediting() on MCeditingimage in setcard() before closing the old card, ensuring the mutable image rep is committed to a serialisable format at navigation time rather than relying solely on recompress() inside MCImage::close().

### DIFF
--- a/engine/src/card.cpp
+++ b/engine/src/card.cpp
@@ -2498,6 +2498,10 @@ MCImage *MCCard::createimage()
 {
 	MCtemplateimage->setrect(rect);
 	MCImage *iptr = new (nothrow) MCImage(*MCtemplateimage);
+	// Assign a proper ID so the MCObjptr can resolve this image after
+	// card close/reopen (clearref() nulls the pointer; getref() re-binds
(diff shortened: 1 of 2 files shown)


emily-elizabeth:
tested on macOS 26.6.1, Windows 11 and Ubuntu 25.04

bwmilby (approved):

bwmilby:
Tested on 26.5.2

PR #428, opened 2026-08-05
380 unable to paste formatted text into field

### LINKED ISSUE #380: Unable to paste formatted text into field
HyperXtalk 0.9.15 running on Mac M2 15.7.7 Sequoia

Prepare a multi line block of text in textedit.  Include text of different colours and make some bold etc.

Open the object inspector on a label field and copy and paste the formatted text into the field.  The result is the cursor in the field moves down a number of lines but no text is displayed.  Note change the background colour of field to confirm that it is not an issue of  font colour.

It is possible to paste in plain text but often it is necessary to add an additional linefeed/cr to the end to get the text to appear.

emily-elizabeth:
Fix RTF paste losing text when pasting formatted text into a field on Mac

Problem

When copying formatted text (different colors, bold, etc.) from TextEdit and pasting into a label field on macOS ARM, the cursor would move down the correct number of lines but no text would appear. Plain text paste worked fine; RTF paste was broken only on Mac.

Root cause

The RTF paste pipeline converts Windows-1252 encoded bytes to UTF-16 via TextConvertToUnicode in dskmac.mm. This function uses a two-pass approach: the first call probes the required buffer size, the caller resizes, then a second call does the actual conversion.

The bug was in dskmac.mm (the Objective-C++ file compiled on Mac — distinct from dskmac.cpp). When the output buffer was too small, the code clipped the output byte count to 0 and returned true, indicating success. The caller interpreted this as a successful zero-byte conversion, advanced by nothing, cleared the input, and moved on — silently discarding all text content. The resize-and-retry path was never reached.

Fix

When the required output size exceeds the available buffer, return false with r_used set to the required byte count. RTFText::Flush() already handles this correctly: it calls Ensure(r_used) to grow the buffer, then retries the conversion. The second call succeeds and the text flows through to the field.

The same fix is applied to dskmac.cpp for consistency, though that file is not compiled on Mac.

Testing

Verified that copying formatted text (multiple colors, bold) from TextEdit and pasting into a field now correctly pastes all text with formatting preserved.

### DIFF
--- a/engine/src/dskmac.cpp
+++ b/engine/src/dskmac.cpp
@@ -3625,13 +3625,18 @@ struct MCMacDesktop: public MCSystemInterface, public MCMacSystemService
         CFIndex t_char_count = CFStringGetLength(t_cf_str);
         CFIndex t_byte_count = t_char_count * (CFIndex)sizeof(UniChar);
         if (t_byte_count > (CFIndex)p_output_length)
-            t_byte_count = (CFIndex)p_output_length;
+        {
+            // Buffer too small — report required size so the caller can resize and retry.
+            r_used = (uint4)t_byte_count;
+            CFRelease(t_cf_str);
+            return false;
+        }
         CFStringGetCharacters(t_cf_str,
-            CFRangeMake(0, t_byte_count / (CFIndex)sizeof(UniChar)),
+            CFRangeMake(0, t_char_count),
             (UniChar *)p_output);
         CFRelease(t_cf_str);
         r_used = (uint4)t_byte_count;
-        
+
         return true;
     }
     
--- a/engine/src/dskmac.cpp
+++ b/engine/src/dskmac.cpp
@@ -3615,8 +3615,11 @@ struct MCMacDesktop: public MCSystemInterface, public MCMacSystemService
         
         // ARM replacement: use CFString for legacy encoding → Unicode
         CFStringEncoding t_cf_enc = (CFStringEncoding)t_encoding;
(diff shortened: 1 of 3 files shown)


bwmilby:
I've been working on getting the test suite working again (`make check-mac`) and all of the RTF tests were failing.  Did not have a chance to look into that yet but this actually corrected those tests too.  Failures now down to 128 from  149 (out of 42769).

emily-elizabeth:
I love (accidentally) fixing bugs 😃

bwmilby (approved):
Tested successfully in IDE.  RTF tests now pass as well.

bwmilby:
I would remove or comment out the test debug output before merging.

emily-elizabeth:
> I would remove or comment out the test debug output before merging.

Too late, I just hit merge and then this message came in..

emily-elizabeth:
I'm not seeing any debugging left-over

```
emily-elizabethhoward@Emily-Elizabeths-MacBook-Pro mac-bin % /Users/emily-elizabethhoward/Developer/HyperXTalk/mac-bin/HyperXTalk.app/Contents/MacOS/HyperXTalk 2>&1 | grep "RTF-"
emily-elizabethhoward@Emily-Elizabeths-MacBook-Pro mac-bin % 
```

bwmilby:
They are in the .cpp file.

emily-elizabeth:
oh, I think those were left because it's not used, but I'll take care of them. But I did find out that we're quitting with a signal 11 so need to find out why

PR #430, opened 2026-08-07
Restore ability to run tests

bwmilby:
Update makefiles for testing:
- Testing uses the engine but requires command line arguments to work.  The IDE build does not pass command line arguments so the makefiles needed to be adjusted to provide path the the standalone build of the engine.

Get `make check-mac` xcode tests working
- `libfoundation` test would not compile due to not being able to find ffi.  Updated gyp file.
- hash test fails when random real returns 0 (not really fixed).  two statements were on a single line, just separated them
- `lextable` would not validate due to case sensitivity.  lower cased letters causing the problem.

Code fixes:
- Correct execution errors header file
- Update `_testlib` to read new format EE and PE

Test fixes:
- Add arm64 processor to MacOS (lcs/core/engine/engine)
- strict compile fixes (lcs/core/engine/put)
- strict compile fixes (lcs/core/field/rtrText)
- urlencode change (lcs/core/filters)
- updates to account for MacOS sandboxing (lcs/core/files/folders)
- Delete lcs/core/engine/password test (n/a for open source)
- correct JSON library branding (lcs/core/engine/extension)
- skip VLC tests (put, controlatloc, group)
- skip tests for tsNet and mergExt (ext-urls)

Documentation fixes:
- notificationPermission - change Value to Returns
- scriptTextSize - remove extra type (command)

OUTSTANDING ISSUES:
- anything that requires VLC crashes the test suite
- there are a large number of errors generated due to documentation issues

Current results on Mac:
```
100 OF 42782 TESTS FAILED
	3 unexpected passes
	41 expected failures
	19 skipped
```


tests/lcs/core/field/rtfText.livecodescript
emily-elizabeth, line 1: on line 69, can we replace "www.livecode.com" with something else?

emily-elizabeth: and line 77

bwmilby: Updated, good catch.

emily-elizabeth (approved):

PR #432, opened 2026-08-07
397 the ide displays large numbers in an array using scientific notation

### LINKED ISSUE #397: The IDE displays large numbers in an array using Scientific Notation
> This is feedback and there may well be a setting that I have missed but while debugging some code I noticed that a large number used as a file pointer is displayed in the IDE using Scientific Notation. This notation implied that there was a loss of precision. However, the true value is stored as expected with full precision and may be displayed if printed using a put statement.

Image

emily-elizabeth:
Fix: Large integers displayed in scientific notation in treeview

When displaying large numbers in the IDE's array inspector, the treeview widget was converting them using LCB's format as string, which uses %g-style formatting internally and switches to scientific notation for large values (e.g. 140234567890123 displayed as 1.40235e+14). The underlying value was always stored correctly — this was purely a display issue.

The fix adds a private IntegerToString handler that detects whole numbers and converts them digit-by-digit using modulo arithmetic, bypassing the formatter entirely. Genuine floating-point values still use the original format as string path.

### DIFF
--- a/extensions/widgets/treeview/treeview.lcb
+++ b/extensions/widgets/treeview/treeview.lcb
@@ -1939,7 +1939,9 @@ private handler calculateMoreElements(in pLimit as Integer, in pLevel as Integer
 				-- notation (e.g. file pointers displayed as 1.4e+14 instead of
 				-- the full value).
 				if the floor of pArray[tKey] = pArray[tKey] then
-					format "%.0f" with pArray[tKey] into tString
+					variable tInt as Integer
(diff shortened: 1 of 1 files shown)


emily-elizabeth:
I don't know how to get it to display in scientific notation intentionally.

bwmilby (approved):
This does work as intended.  I'm not sure if it is the most efficient/best solution, but I would not hold up progress waiting for a better option.  If a better way is proposed later we can always change it at that time.

PR #433, opened 2026-08-07
[[Issue 431]] Fix widget export crash

### LINKED ISSUE #431: widget export with missing OnSave causes crash
If you attempt to `export widget 1 as array tArray` on a widget that is missing the `onSave` handler it will crash the IDE with a signal 11.

[ColorSwatchWidget.hyperxtalk.zip](https://github.com/user-attachments/files/30830896/ColorSwatchWidget.hyperxtalk.zip)

This bug also impacts LiveCode 10 & 11

bwmilby:
Check for missing `OnSave` handler caused crash when exporting a widget to an array.

Closes #431

### DIFF
--- a/engine/src/widget.cpp
+++ b/engine/src/widget.cpp
@@ -1121,8 +1121,9 @@ void MCWidget::GetState(MCExecContext& ctxt, MCArrayRef& r_state)
     }
     else
     {
-        if (!MCWidgetOnSave(m_widget,
-                            &t_value))
+        if (!MCWidgetOnSave(m_widget, &t_value) || 
+            // A widget might not have an OnSave handler (e.g. colorswatch widget)
+            (*t_value == nil))
         {
             r_state = MCValueRetain(kMCEmptyArray);
             return;


bwmilby:
Line 895 (save) and 945 (clone) also use `MCWidgetOnSave` and check for a `nil` value on the pointer (`(*t_rep == nil)`) preventing the type of crash that we see in the `GetState` function.

emily-elizabeth:
When using your test stack, it's still crashing on macOS 26.6.1 using HyperXTalk 0.9.16 (15863)

bwmilby:
Strange.  I’ll start with config and then compile to retest.

bwmilby:
If I use my `hxsa` alias to use the standalone build, it will crash.  If I build an app with the stack, it does not crash.  In the IDE it is fine.  I'm currently building on Linux to test there.

emily-elizabeth:
Does not crash on Windows 11 and Ubuntu 25.04 after applying this PR

bwmilby:
Just checked on Ubuntu 24.04.4 as well.  Built standalone is fine but using `hxsa` alias does not work out since the widget isn't loaded.

emily-elizabeth:
I made a couple of updates in the save project PR https://github.com/emily-elizabeth/HyperXTalk/pull/443

bwmilby:
I think my way is better.  Don't need both though.

emily-elizabeth:
I don't see any new commits and the last time I tried it it crashed on macOS 26.6.1. Let me know if I missed something.

bwmilby:
The code in this PR does the exact same thing that your other change made.  It only does it in the same way that it is done elsewhere in the code.  I suspect that something was different in your test for some reason.  There should be no difference based on OS and it works everywhere else for you and also for me on Mac.

bwmilby:
This is the way this particular thing is handled elsewhere in the code base.  When you have an `if` statement, you chain the other conditions of failure using `||`.  Once one item is `TRUE` then the rest are not evaluated.  In this case, the save returns successfully, but the pointer is NULL.
Screenshot 2026-08-11 at 10 36 38 PM
This is functionally equivalent, but longer and not in the style of the rest of the code base.  There is no reason that this would not crash and the above would crash since they do the same thing.
Screenshot 2026-08-11 at 10 37 14 PM

emily-elizabeth:
I'll re-run it again Wednesday August 12th and see what happens.

bwmilby:
Sounds good.  I hope I'm not coming across poorly.  I just happen to be a bit frank, hopefully not too much so.  We appreciate your extensive contributions to the open source community.

emily-elizabeth:
> Sounds good. I hope I'm not coming across poorly. I just happen to be a bit frank, hopefully not too much so. We appreciate your extensive contributions to the open source community.

I just feel bad that you real programmers have to work with a "hack" like me 😞

emily-elizabeth (approved):

bwmilby:
Looks like a merge added some extra code to widget.cpp on the project branch.  I could push a commit to remove the extra lines if you like.

emily-elizabeth:
Yes please. I'll be out for a bit so feel free to fix anything you see.

PR #442, opened 2026-08-09
fix(splash): centre splash screen on IDE's monitor at launch

### LINKED ISSUE #378: linux multiple monitors issues
On my system with three monitors: laptop in the middle and primary screen, one to the left, one to the right:

Splash screen is always on the laptop, even though the menubar etc are by desire on screen 1.
The standalone builder progress window is also on the laptop screen.

When using the dropper tool the screen size is wrong: seems like the default screen size is just the laptop screen size. The cursor changes from the dropper cursor to the default cursor when getting past those borders, even though the dropper tool works properly.

emily-elizabeth:
fix(splash): centre splash screen on IDE's monitor at launch

Problem

On multi-monitor setups where the HyperXTalk launcher sits on a different monitor than the IDE's saved window positions, the splash screen always appeared on the primary (laptop) monitor while the IDE itself opened on the external monitor.

The root cause was that PreOpenStack in the splash behaviour fires before the behaviour stack is loaded into memory, so any positioning code placed there was silently ignored. The original set the location of this stack to the screenLoc was never executing — the splash was simply opening at its last saved position (centre of the primary screen).

Fix

Move the positioning logic to setProp statusMessage, which fires on the first IDE startup status update — by which point the behaviour is definitely active.

On each call, the handler runs only once (guarded by a script-local sPositioned flag) and delegates to revSplashBehavior__CentreOnIDEScreen, which:

Reads the saved palette_rect_revMenubar property from the preferences stack (~/.hyperxtalk/preferences/hyperxtalk7.rev) to determine which screen the IDE's menubar was last on
Walks the screenRects to find the screen containing the centre of that rect
Centres the splash on that screen

Falls back to the mouse cursor position on first run (when no preferences file exists yet).

Files changed

ide/Toolset/palettes/splash/revsplashstackbehavior.livecodescript

### DIFF
--- a/ide/Toolset/palettes/splash/revsplashstackbehavior.livecodescript
+++ b/ide/Toolset/palettes/splash/revsplashstackbehavior.livecodescript
@@ -1,45 +1,33 @@
 ﻿script "revSplashStackBehavior"
 setProp statusMessage pMessage
    put pMessage into field "status" of me
-   -- Once the menubar stack is open we know which screen the IDE is on.
-   -- Reposition the splash there (no-op after the first successful snap).
-   if exists(stack "revMenubar") then
-      revSplashBehavior__CentreOnIDEScreen
-   end if
 end statusMessage
-
-private command revSplashBehavior__CentreOnIDEScreen
-   local tRefLoc, tRefH, tRefV
-   put the loc of stack "revMenubar" into tRefLoc
-   put item 1 of tRefLoc into tRefH
-   put item 2 of tRefLoc into tRefV
-
-   local tTargetRect
-   put line 1 of the screenRects into tTargetRect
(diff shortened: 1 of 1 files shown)


emily-elizabeth:
tested on Ubuntu 25.04

emily-elizabeth:
tested on Linux Mint, Debian 13.5 (GNOME and MATE) and Fedora

emily-elizabeth:
tested on Windows 11, but the splash screen always opened on the main monitor and not on the external, so will have to look into that

emily-elizabeth:
tested on Windows 11, had to change some code as it was hardcoded for Linux preference file, so now it uses the function from home.livecodescript to get the preference path

PR #443, opened 2026-08-09
355 save as project issues

### LINKED ISSUE #355: Save as project issues
The new Save as project feature has a few issues that need to be addressed.  I'll use this issue to collect them - some have been identified elsewhere as the issue impacts more than just this feature.

- Unplaced backgrounds/shared groups are not exported.
- Save overwrite message incorrect.  Line 8167 of revidelibrary calls `revIDELocalize` incorrectly.  The second parameter is expected to be a list (numerically keyed array starting at 1).
- Widgets are not fully exported (rect, etc) due to `the properties` returning an empty list.  This issue impacts more than just this feature and is identified in issue #341 .
- Widget state is not exported/restored.  This requires an additional step after obtaining the basic properties (mentioned above).  `export widget x to array y` needs to be used to capture the state.  The kind is currently exported/restored but would also be captured with this command.
- When exporting custom properties the possibility exists for binary data.  The current JSON implementation creates invalid JSON as it does not escape those values.  Either the library needs to encode binary or it should report an error so the data can be encoded before passing to the JSON library.  The method of correction here will depend on how #343 is addressed.  Issue may be related to how `revIDEProjectSanitizeArray` works since it forces the leaf nodes to be coerced to string values.  The code in the JSON library cannot handle strings with null values since it uses null terminated strings.
- JSON is currently exported in a format intended for data exchange (a single line).  This is useful for data transfer but does not work well for human review.  My assumption is that one of the reasons for the project feature is for version control. If that is the case, then Git will be much more useful with a pretty print version of the JSON file.
- Using the same extension for the JSON as the project is confusing.  My suggestion would be to use a distinct extension for the file inside the folder.  I would suggest `stackName.json`
- Currently the scripts for the stack are exported as a part of the JSON.  Editing the scripts with an external editor could be facilitated fairly easily if all of the scripts lived in a single folder inside the project package.  My scriptTracker plugin could be reviewed to see the concept that I had used.  Even without facilitating external editors, version control would be much easier to track if the scripts were exported per object with either names that linked back to the object that was understandable by a human or if metadata was included in the file that linked it.  I used both for scriptTracker.
- Another system (SmartCrumbs) splits all text (or is it just htmltext?) out to individual files.  I'm not sure if going that far is necessary or even beneficial, but it is something that could possibly be considered.

Since I mentioned scriptTracker a couple times, here is a link:
https://github.com/bwmilby/scriptTracker

emily-elizabeth:
Fix project file format: items 5–10

This PR completes the known gap list for the .hyperxtalk_project format, building on the earlier items 1–4.

Item 5 — Button icon IDs
Button icon properties (icon, hilitedIcon, disabledIcon, visitedIcon, armedIcon) are numeric object IDs that change every time a stack is reconstructed. The serializer now strips these IDs from the properties blob, looks up the referenced image's short name, and stores it in an iconNames map. On load, icon resolution is deferred (same pattern as behaviors) until all controls exist, then each name is resolved back to the new ID.

Item 6 — Behavior resolution warning
Previously, if a behavior button couldn't be found after load (missing stack, renamed button), the failure was silent. The reconstructor now collects all unresolved behaviors and shows a single warning dialog after load listing each one by name and stack, with guidance to check that all behavior stacks are included in the project.

Item 7 — Pretty-printed JSON
The project JSON is now written with jsonExport(tTree, true), producing indented, human-readable output that produces meaningful diffs in version control.

Item 8 — revIDEMoveSubstackToFile
This handler was a complete stub. It now validates that the target is a substack (not already a main stack), prompts the user for a destination .hyperxtalk file, and calls set the filename to detach the substack from its parent and save it as an independent file.

Item 9 — Painted card graphics
Investigation of the engine (MCCard::createimage in card.cpp) confirmed that the LiveCode paint tools do not maintain a separate paint layer — they create ordinary MCImage controls on the card. These are already fully handled by the existing control serializer, so no additional work was needed.

Item 10 — cRevProjectFilePath re-serialization
The project bundle path stored in cRevProjectFilePath was being round-tripped into the JSON on every save, meaning it would become stale whenever the project was moved or renamed. It is now stripped from the custom properties array before export. The property continues to be set at load time from the actual bundle location.


bwmilby:
One general comment... I don't think maintaining backward compatibility for pre-release versions of a feature is necessary.

emily-elizabeth:
oh, I'm not 😉

bwmilby:
Couple of shortcuts you could employ:
- `-1` is the index of the last char, so you can say `char 12 to -1 of tString` instead of manually calculating length (found in binary decode)
- you can safely pass an array element by reference without assigning it to a variable first:  `revIDEProjectDecodeBinaryProps pProps[tKey]`

Changes are looking good.

emily-elizabeth:
> * `-1` is the index of the last char, so you can say `char 12 to -1 of tString` instead of manually calculating length (found in binary decode)

yeah, Claude thought it wasn't allowed and when I corrected him, he was lazy about it and just kept that code *lol*

bwmilby:
> > * `-1` is the index of the last char, so you can say `char 12 to -1 of tString` instead of manually calculating length (found in binary decode)
> 
> yeah, Claude thought it wasn't allowed and when I corrected him, he was lazy about it and just kept that code _lol_

I think you need to make Claude fix it.  It is poor xTalk programming to do it that way.  Not just more words, but it is inefficient and slower.  Won't make much difference in performance here, but not something he should get away with.

bwmilby:
Why are you limiting `scalefactor` to >1?  It is a valid thing to want to downscale a larger stack to see more of it.

emily-elizabeth:
> Why are you limiting `scalefactor` to >1? It is a valid thing to want to downscale a larger stack to see more of it.

I added `AND tScaleFactor > 0` so even scale factor 0.5 still works. according to the docs you can't go under 0 scale factor and the fact that the next line was throwing an error because the tScaleFactor = 0 I had to fix it.

bwmilby:
My eyes are playing tricks on me 😉

emily-elizabeth:
> My eyes are playing tricks on me 😉

No worries, we old people have that problem. Imagine if I didn't wear my glasses how much worse the code would be 😄

bwmilby:
`jsonImport` does not zero the value for a `borderWidth` key

emily-elizabeth:
> `jsonImport` does not zero the value for a `borderWidth` key

The borderWidth was exported out to the JSON file correctly (borderWidth = 1). When reading it back in, tJson had the correct borderWidth property but right after the JsonImport tTree had all zeroed out borderWidth.

emily-elizabeth:
> I just did a test with a file that has it in there. It survived the import. Can you attach the json file?

[datagridTest.hyperxtalk_project.zip](https://github.com/user-attachments/files/30915491/datagridTest.hyperxtalk_project.zip)

bwmilby:
Unquoted integers are being converted to 0 upon import.  This issue is unrelated to the key name, it is an issue with numeric values.  Numbers that contain a decimal point are imported (1.0 becomes 1, 1.1 becomes 1.1).

Issue is that we are using `yyjson_get_raw` incorrectly.  This means that we are always getting a real.  Since integers are not real, 0 is being returned.

bwmilby:
PR #446 corrects the integer import issue.

bwmilby:
There is a simpler way to create your files.  Here is an example from one of my scripts:
```
put textEncode(tScript, "UTF-8") into url ("binfile:" & tFileName)
```

bwmilby:
I just tried my "TortureTest" stack and things looked mostly right when opened except for the segmented widget.
```
errorDialog "863,0,0,runtime
864,0,0,Value is not of correct type for assignment to variable - expected type <type: livecode.lang.list> for assigning to variable mSegmentNames in com.livecode.widget.segmented.
865,0,0,segmented.lcb
866,0,0,254
897,0,0,1"
8:58:46 PM
```
Screenshot 2026-08-12 at 8 59 58 PM

I'm going to categorize this as a bug in the widget over a bug in the new feature.  Issue #447 

There is an issue with the JSON library though.  Objects are being exported as dictionaries (LCB array) that should be arrays (LCB lists).  Another issue is that dictionaries should probably be exported with keys sorted otherwise tracking changes in a VCS becomes more difficult since the values could jump around.

bwmilby:
The segmented fix works, but really should not be in this PR as the issue is not related to the new feature.  I also solved it (just not published) in a similar way but did not restrict the input array to a properly formatted LCS list (sorted the keys and used that to build the LCB list).  The issue I posted checked and determined that we can't round trip a subkey list (so `arrayencode` would not help) and the issue remains in LC11 even with its better support for differentiating between arrays and lists.  I've not looked for an instance where just a list is returned to see if that can round trip.

bwmilby:
I'll work on a validation test.  I'm going to build a stack with every widget and some test data in each.  Method will be to save as binary, do an export with another tool, save as project, quit & relaunch, open project, export with another tool, compare results.  Except for object IDs, results should be the same.

bwmilby:
Providing a diff of the JSON was not easy.  Due to how the file is reconstructed, the shared groups change location.  I have found a several issues though:
* when `sharedText` is false (default value), differing text values from other cards is not captured in shared groups [pg 39/41]
Screenshot 2026-08-16 at 5 26 07 PM

* Data Grid Template stack names lose their spaces in some places (original file / project version) [pg 17/18]
  * "left swipe control": "group id 1016 of card id 1005 of stack \"Data Grid Templates 1781228370148\""
  * "left swipe control": "group id 1016 of card id 1005 of stack \"DataGridTemplates1781228370148\""  
Screenshot 2026-08-16 at 5 23 21 PM

* non-background shared groups are being copied to each card vice being placed [pg 23/37]
Screenshot 2026-08-16 at 5 24 59 PM

* unplaced groups to not survive [pg 41/42]
Screenshot 2026-08-16 at 5 20 48 PM

* objects without a specific name have the name set to the form `widget id 1018` using original id [pg 19]
Screenshot 2026-08-16 at 5 33 23 PM

Page numbers are from the PDF in the next comment.

bwmilby:
[TortureTest.json and TortureTest.json - Deltopia DeltaWalker.pdf](https://github.com/user-attachments/files/31074300/TortureTest.json.and.TortureTest.json.-.Deltopia.DeltaWalker.pdf)
Here is a full diff of the two exports.  I manually reordered some of the keys so the diff would be more useful.

What we really need is a binary stack that is constructed in such a way that when saved as a project that it is opened/built with the same IDs.  The last 3 issues in the previous comment need to be corrected before that will be possible though.

One other thing I just noticed is that the one image that I had in the stack changed size.  Visually it looks pretty much the same, but it was 6.64KB in the original stack and the project version was 3.15KB.  My tool (DeltaWalker) doesn't try to figure out file name changes which is part of the reason for my above comment.  If the order doesn't change, comparing the JSON is pretty easy since the IDs can be easily skipped when looking for "real" changes.

I suppose that is would be helpful to make the stack that I'm using as a test available:
[TortureTest.hyperxtalk.zip](https://github.com/user-attachments/files/31123705/TortureTest.hyperxtalk.zip)

emily-elizabeth:
Sorry, I can't read the black text on the dark green, just not enough contrast for these old eyes.

bwmilby:
Don't really need to read it.  The large green blocks are going to be either the duplicated groups or the missing groups.  Most of the small one line ones are just the IDs which are expected to be different.  The comment with the page number references should have enough detail.  It doesn't look like I can change colors on the PDF output (at least not easily - there is a color section but it doesn't really look to map to the print).  I'm not sure how much better orange on black is, but I've updated the original comment with screen shots of pieces of the diff.

bwmilby:
I've updated my 3 earlier comments.  I added screen shots to the initial report.  I added a note about the difference in image size.  I attached the stack that I used to test the build.

emily-elizabeth:
> Data Grid Template stack names lose their spaces in some places (original file / project version) [pg 17/18]
> "left swipe control": "group id 1016 of card id 1005 of stack "Data Grid Templates 1781228370148""
> "left swipe control": "group id 1016 of card id 1005 of stack "DataGridTemplates1781228370148""

This is due to a bug in the JSON pretty print, it has finally been worked around

emily-elizabeth:
> objects without a specific name have the name set to the form widget id 1018 using original id

we set the control back to the original ID so this issue should be taken care of

emily-elizabeth:
> when sharedText is false (default value), differing text values from other cards is not captured in shared groups [pg 39/41]

I don't understand what you mean by this

bwmilby:
> > objects without a specific name have the name set to the form widget id 1018 using original id
> 
> we set the control back to the original ID so this issue should be taken care of

The problem is that the name is actually set to the text "widget id 1018".  That is how the name is presented if one is not set.  The actual name should remain blank/unset (I think).
Screenshot 2026-08-21 at 8 32 37 AM

bwmilby:
> > when sharedText is false (default value), differing text values from other cards is not captured in shared groups [pg 39/41]
> 
> I don't understand what you mean by this

Download the stack that I posted to see an example.  There are a couple of fields in shared groups where "sharedText" is false and one where it is true.  When it is false, each card where the field appears will have a unique value for the field.  It is only when true that the value will be the same on all cards.  Those groups are on cards 1 & 2.
Screenshot 2026-08-21 at 9 01 29 AM
Screenshot 2026-08-21 at 9 01 43 AM

bwmilby:
The lists of background groups and shared groups are not necessarily the same.  There is a separate option to make a group behave as a background (changes message path).  The project is capturing a list of background group names per card, but that is problematic.  If multiple backgrounds have the same name, there is no way to ensure the correct one is being referenced.  Shared/background groups should be captured at the stack level like cards.  Shared groups are currently being exported once per card.  Background groups do not appear to be exported.  Unplaced groups are also ignored (I try to get them in BST, but I still have some where I'm not getting the controls properly).

emily-elizabeth:
> The problem is that the name is actually set to the text "widget id 1018". That is how the name is presented if one is not set. The actual name should remain blank/unset (I think).

An object has to have a name set, hence the reason why the IDE auto-names them with "widget id 1018" so either set a name or deal with the auto-name

bwmilby:
> > The problem is that the name is actually set to the text "widget id 1018". That is how the name is presented if one is not set. The actual name should remain blank/unset (I think).
> 
> An object has to have a name set, hence the reason why the IDE auto-names them with "widget id 1018" so either set a name or deal with the auto-name

Not quite.  You always get a value when you ask for the name, but it does not need to be set.
```
put the name of group id 1005
--> group id 1005
set the name of group id 1005 to "temp"
put the name of group id 1005
--> group "temp"
set the name of group id 1005 to empty
put the name of group id 1005
--> group id 1005
set the name of group id 1005 to "group id 1005" -- this is what is happening
--> group "group id 1005"
```

emily-elizabeth:
> ```
> set the name of group id 1005 to "group id 1005" -- this is what is happening
> --> group "group id 1005"
> ```

which is what happens in the IDE with binary stacks, so the issue lies outside of the project format.

bwmilby:
> > ```
> > set the name of group id 1005 to "group id 1005" -- this is what project format is doing
> > --> group "group id 1005"
> > ```
> 
> which is what happens in the IDE with binary stacks, so the issue lies outside of the project format.

The project import is setting the name incorrectly in this instance.  The IDE never sets the name of an object like that by default.  When you import an object, if the name is in the format "objectType id number" then you need to leave the name empty.  Setting the name results in the double object format that I'm talking about.  My point is that you can't set the default "id" format name of an object, the engine just reports it out that way.

**group id 1005** is not the same as **group "group id 1005"**

emily-elizabeth:
> When you import an object, if the name is in the format "objectType id number" then you need to leave the name empty. 

gotcha. except for stacks (as per the documentation on `name`)

bwmilby:
True.  Stacks do always start out as "Untitled" which has its own issue :)

emily-elizabeth:
> True. Stacks do always start out as "Untitled" which has its own issue :)

What's the issue?

bwmilby:
Unique stack name requirement.  If you forget to rename a stack, it is easy to end up with conflicts.

emily-elizabeth:
> Unique stack name requirement. If you forget to rename a stack, it is easy to end up with conflicts.

No, they use a sequential numbering to name the stacks "Untitled 1", "Untitled 2", etc

bwmilby:
In a single session, yes they do have numbers.  But easy to have a bunch of "Untitled 1" stacks around.  I try to name everything before I save it, but the unchanged ones linger on my various folders all over the place.

emily-elizabeth:
Yes, I've been bitten by that too, but am getting better at naming the stacks something else. Perhaps we should change it to Untitled && the milliseconds?

emily-elizabeth:
> > > when sharedText is false (default value), differing text values from other cards is not captured in shared groups [pg 39/41]
> > 
> > 
> > I don't understand what you mean by this
> 
> Download the stack that I posted to see an example. There are a couple of fields in shared groups where "sharedText" is false and one where it is true. When it is false, each card where the field appears will have a unique value for the field. It is only when true that the value will be the same on all cards. Those groups are on cards 1 & 2. Screenshot 2026-08-21 at 9 01 29 AM Screenshot 2026-08-21 at 9 01 43 AM

I think this issue has been resolved.

emily-elizabeth:
> The project import is setting the name incorrectly in this instance. The IDE never sets the name of an object like that by default. When you import an object, if the name is in the format "objectType id number" then you need to leave the name empty. Setting the name results in the double object format that I'm talking about. My point is that you can't set the default "id" format name of an object, the engine just reports it out that way.
> 
> **group id 1005** is not the same as **group "group id 1005"**

This has been fixed

emily-elizabeth:
> In a single session, yes they do have numbers. But easy to have a bunch of "Untitled 1" stacks around. I try to name everything before I save it, but the unchanged ones linger on my various folders all over the place.

https://github.com/emily-elizabeth/HyperXTalk/issues/455

bwmilby:
> Yes, I've been bitten by that too, but am getting better at naming the stacks something else. Perhaps we should change it to Untitled && the milliseconds?

That would certainly work.  Somewhere I've seen where a change was made (plugin maybe) where when you created a new stack, it asked you for a name (it had something filled in as a default).  Seconds would be unique enough.

emily-elizabeth:
Brian, think you can check this to make sure all the changes are made and made correctly?

bwmilby:
Yes, I'll update and run through my test stack to ensure everything tracks.  Should be able to run through tonight.

bwmilby:
- Shared and non-shared text restored properly.
- Objects with no name set are importing correctly.
- DG stack name importing correctly (retaining spaces).
- There is an issue restoring the "background behavior" flag on background groups.  The shared flag is being applied, but not the background.  The property value is stored in the project's JSON file though.
- Got an error on DG behaviors:
Screenshot 2026-08-23 at 8 26 25 PM

Here is what I'm seeing in the project browser before:
Screenshot 2026-08-23 at 8 59 55 PM

And after:
Screenshot 2026-08-23 at 9 00 42 PM
So the issue is in the Row Template.  When I look at the group in the project JSON, I only see a stack reference for the behavior and not the button: `"behaviorStack": "Data Grid Templates 1781228370148"` (not stored as a property).  I didn't see a separate key for the button.  In my file, I store it as a property:
```
"behavior": "button id 1009 of stack \"Data Grid Templates 1781228370148\""
```

The most confusing thing to me is the big difference in shared groups between the original (top) and project:
Screenshot 2026-08-23 at 9 16 07 PM
I can upload my 2 JSON exports if you would like to take a look at them.  My structure is different than yours, but you should be able to follow it.

Changing the JSON library to sort array keys before emitting them would make comparing much easier (and for version control will be important).  I use a slightly custom version of PhotonJson in my tool.  That is one of the things I added.

emily-elizabeth:
Feel like checking the changes?

bwmilby:
Yes, I'll take a look later this evening.

bwmilby:
I think the main stack itself made it back and forth mostly in tact.  The DG stack ended up with all of the non-background groups on the second card duplicated with one placed on the card and one unplaced.  I can't tonight, but tomorrow I may be able to peek at the code and see if I notice anything obvious.

emily-elizabeth:
> to peek at the code and see if I notice anything obvious.

Any chance you've looked to see anything?

bwmilby:
I have been side tracked by another related project and have not looked yet.  I'm going to take a look now.

emily-elizabeth:
No worries. When you have time 😄

bwmilby:
Nothing jumped out when looking last night but the rebuild code is quite involved.  I think I need to do a non-DG substack test since that simplifies the code flow.

emily-elizabeth:
Are we moving forward with this or is it good to be reviewed and merged?

bwmilby:
The current state of this PR is certainly better than what is on main.  I would still consider it alpha/beta quality though.  There are other things mixed in this PR that probably need to be in main (widget load fix, json fix).  The group duplication is the thing that concerns me a bit since it could lead to corruption that just gets worse with every save/load cycle - I didn't think to test multiple iterations of load/save before to see what actually happens though.

When I looked earlier, the group issue did not seem to be the exported data, but in the way it was reconstructed.  The group duplication didn't exist in the json export file.  I also noticed that the layer order seemed to be reversed on import.

PR #448, opened 2026-08-13
Feature/markdowntext

emily-elizabeth:
Add markdownText field property

Adds markdownText as a first-class field property alongside the existing htmlText and rtfText, allowing field content to be set and retrieved as CommonMark-subset Markdown.

What's supported

On import (setter):

ATX and setext headings (# through ######, ===/--- underlines) — sized to match the HTML importer (34/24/18/14/12/10pt, H5 italic, H6 italic)
Unordered lists (-, *, +) and ordered lists (N.)
Fenced code blocks (` or ~) and indented code blocks
Blockquotes (>)
Horizontal rules (--- / *** / ___)
Inline: ***bold-italic***, **bold**, *italic*, `code span` (Courier), [text](url) links, backslash escapes

On export (getter):

Headings detected by paragraph metadata or font size → # prefix
Bold/italic/bold-italic → **/*/***
Code span font → backticks
Links → [text](url)
Lists → - / N.
Blockquote indent → > prefix

Files changed

engine/src/fieldmarkdown.cpp — new file; all import and export logic
engine/src/field.h / fieldh.cpp — setmarkdown() and method declarations
engine/src/exec-interface-field.cpp / exec-interface-field-chunk.cpp — Get/SetMarkdownText and chunk variants
engine/src/exec-interface-object.cpp — property name tables
engine/src/parsedef.h — P_MARKDOWN_TEXT constant
engine/src/lextable.cpp — markdowntext keyword (alphabetical)
engine/src/mcstring.h / mcstring.cpp — MCN_markdowntext name ref
engine/engine-sources.gypi — fieldmarkdown.cpp added to build

Bugs fixed during development

Infinite loop on [ with no following ](url) (e.g. [ ] checkboxes, [Topic] headings)
Infinite loop on unmatched */_ caused by uint32_t wrapping past zero
UTF-8 multi-byte characters (en dash, em dash, curly quotes, etc.) decoded as MacRoman — fixed by splitting at the ASCII boundary and passing non-ASCII runs through MCStringCreateWithBytes(..., kMCStringEncodingUTF8) → UTF-16 with is_unicode = true
Consecutive source lines (e.g. table rows) concatenated into a single field paragraph with no separator — fixed by ending the paragraph after every regular line


emily-elizabeth:
tested on macOS 26.6.1 using the samples from https://markdown.co.in/examples.html

emily-elizabeth:
tested on Windows 11

emily-elizabeth:
tested on Ubuntu 25.04

emily-elizabeth:
[markdownTest.hyperxtalk.zip](https://github.com/user-attachments/files/31049749/markdownTest.hyperxtalk.zip)

bwmilby:
Tested on Mac 26.6.1 as well.  Seems to do the things listed in the initial post.  Only obvious thing missing from the examples is table handling.

emily-elizabeth:
Does the field object support tables?

bwmilby:
It does not have good support for inline tables.  There is a table field, but the tab stops are for the entire field.

bwmilby (approved):

PR #450, opened 2026-08-17
Feature/frame image data

emily-elizabeth:
Add frameText, frameImageData, and frameDuration properties to image

This PR exposes three new read-only properties on image objects to support non-destructive access to individual frames of animated GIFs.

New properties

frameText — Returns the current frame encoded as PNG data. Suitable for passing directly to LCB widgets via a binary data property or setting on another image object with set the text of image. This is the primary property for driving animated GIF playback through an external widget.

frameImageData — Returns the raw composited ARGB pixel data for the current frame, in the same format as imageData. Accessing this property does not alter currentFrame or trigger a repaint, making it safe to use during active animation.

frameDuration — Returns the display duration of the current frame in milliseconds, as encoded in the GIF file. Use this to drive a correctly-timed animation loop.

Usage

Set currentFrame to select a frame, then read the desired property:

livecode
set the currentFrame of image "myGIF" to tFrame
set the viewerImage of widget "imageViewer" to the frameText of image "myGIF"
send "nextFrame" to me in (the frameDuration of image "myGIF") milliseconds

Implementation notes

All three properties use m_rep->LockBitmap(currentframe, 1.0, t_bitmap) to access the decoded frame directly without side effects. frameText encodes the bitmap to PNG using the existing MCImageCreateClipboardData utility. For non-animated images, each property falls back to its existing equivalent (text, imageData, and 0 respectively).

Documentation added for all three properties in docs/dictionary/property/.


bwmilby:
Code looks reasonable.  I'll pull it down and build/test this evening.

bwmilby (changes requested):
Updated the docs for `frameText` so the code block displayed correctly.

I notice that whenever I `get frameText` or `get frameImageData` that the animation stops.  If I set the `repeatCount` to -1 again then it will resume.  I'm not sure the best way to fix it though.  

I think this is the issue (part of `closeimage`):
```
	// cancel any internal messages for this object (used for GIF animation)
	MCscreen->cancelmessageobject(this, MCM_internal);
```
This mean that we either need to avoid `closeImage` (but would also need to avoid the `openImage` due to the reference counting) or we need to reschedule the animation message.

bwmilby:
I guess, based on the presented use case, it could also be documented that it will halt animation of the image.

emily-elizabeth:
> I notice that whenever I `get frameText` or `get frameImageData` that the animation stops. If I set the `repeatCount` to -1 again then it will resume. I'm not sure the best way to fix it though.

Yeah, it was meant to step through the frames and get the data from the frame, not to grab the frameText/ImageData while playing the animated GIF

bwmilby:
When you set the `currentframe` it does not stop the animation.  Probably enough to add a note to frameText and frameData that animation will stop when getting the data and needs to be restarted once done if desired.  I'll let you wordsmith it though :)

emily-elizabeth:
> When you set the `currentframe` it does not stop the animation. Probably enough to add a note to frameText and frameData that animation will stop when getting the data and needs to be restarted once done if desired. I'll let you wordsmith it though :)

Note has been added to the documentation

bwmilby (approved):

PR #457, opened 2026-08-23
creates the new stack with the name "Untitled" && the seconds

### LINKED ISSUE #455: naming of new untitled stacks
when creating a new stack, it is named "Untitled 1", "Untitled 2", etc but if you quit HyperXTalk and relaunch, then numbering starts all over again. Perhaps we can save the last used number in the preferences and keep it moving forward from there

emily-elizabeth:
creates the new stack with the name "Untitled" && the seconds

### DIFF
--- a/ide/Toolset/libraries/revidelibrary.8.livecodescript
+++ b/ide/Toolset/libraries/revidelibrary.8.livecodescript
@@ -6761,17 +6761,18 @@ Tags: ide
 */
 on revIDENewMainstack pWidth, pHeight
    # Set the width and height of the template stack
-
+   
    # Generate a unique name for the stack
-   local tNumber
-   put 1 into tNumber
-   repeat
-      if there is a stack ("Untitled" && tNumber) then  add 1 to tNumber
-      else exit repeat
-   end repeat
-
+   --   local tNumber
+   --   put 1 into tNumber
+   --   repeat
+   --      if there is a stack ("Untitled" && tNumber) then  add 1 to tNumber
+   --      else exit repeat
+   --   end repeat
+   
    # Setup the template stack
-   set the name of the templateStack to "Untitled" && tNumber
+   --   set the name of the templateStack to "Untitled" && tNumber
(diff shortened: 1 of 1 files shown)


bwmilby (approved):
Tested on my Mac and works as expected.

emily-elizabeth:
I wonder if this should be made as a preference? (see the conversation on the original ticket)

ide/Toolset/libraries/revidelibrary.8.livecodescript
```diff
@@ -8039,6 +8039,19 @@ on revIDEActionSaveStackAs pStackID, pSubstackOnly
...
    put the result into tVersion
    
+   // get the saved file name from the dialog
+   local tStackNameFromFileName
+   set the itemDelimiter to "/"
+   put item -1 of it into tStackNameFromFileName
+   set the itemDelimiter to "."
+   delete item -1 of tStackNameFromFileName
+   
+   if (tShortName begins with "Untitled") OR (tShortName <> tStackNameFromFileName) then
```
bwmilby, line 8048: I think that you meant "AND" - we don't want to force the stack name and file name to match in all cases I don't think.  This change is meant to alleviate the situation where the developer forgets to change the stack name of a new stack before saving it.

emily-elizabeth: It's in the Save As, so the OR is meant to change the stack name if the end user save as and changes the file name of the stack we want the stack name to update with it, otherwise we end up in the same situation of having stack names being duplicated.

bwmilby: Ok.  I don't think we want to do both.  If we are going to change the name like this, then revert the seconds piece.

emily-elizabeth: Will do. Will revert it back to Untitled 1, Untitled 2, etc

emily-elizabeth:
Restored the original naming sequence - Untitled 1, Untitled 2, etc

PR #459, opened 2026-08-23
449 Add WebP image format support

### LINKED ISSUE #449: Support WebP image format
Support WebP image format

emily-elizabeth:
- Add iwebp.cpp codec using libwebp 1.4.0 (decode and encode)
- Vendor libwebp 1.4.0 into thirdparty/libwebp/
- Wire WebP into import/export pipelines and lextable
- Add 'webp' format to clipboard MIME types on all platforms
- Add webp to IDE file picker extension list
- Add libwebp to GYP build and prebuilt/build-libraries.sh for all platforms


emily-elizabeth:
Test WebP image, from Google https://developers.google.com/speed/webp/gallery1

emily-elizabeth:
Animated WebP from https://convertico.com/samples/webp/

tom-jerry-animated-webp

emily-elizabeth:
verified on Windows 11

emily-elizabeth:
verified on Ubuntu 25.04

bwmilby:
I did a test and images in webp format displayed correctly.  I set the repeat to -1 on an animation but that value did not save.  When opening the stack, the animation had to be restarted.  Is that the expected behavior?  GIF images retain the repeat -1 value.

emily-elizabeth:
I'll take a look to see about saving the repeat value

emily-elizabeth:
I'm not seeing this issue.
- launch HXT
- import the animated WebP (the Tom and Jerry from above)
- (it's animated and looping)
- save the stack
- relaunch HXT
- open the stack
- (animation is playing)

bwmilby:
I added an image control from the tool bar and then selected the image source.  The animation does not start automatically.  If I import a webp file as an image, then the animation does work correctly..

emily-elizabeth:
- launch HXT
- add image control from the Tools palette
- set the filename of the image to an animated WebP/GIF image
- set the repeat count to -1
- save the stack
- close and remove from memory
- reopen the stack
- animated image does not play (regardless if it is a WebP/GIF image)

Are we looking to change the behaviour of both formats to start playing when opening the stack? Right now the WebP animated image mimics the behaviour of animated GIF images

bwmilby:
That is very interesting.  You are correct - GIF and WEBP animations behave the same.  Both are inconsistent depending on how created.  If you import the control, then the repeat is set to -1 and persists (you can freely change it and it will restore when reloaded).  If you create an image control and set the file, then the repeat value does not restore (can't say for sure what is in the file since it would need to be decoded).  As such, this isn't the place to address it.  The WEBP addition should be separate from any discussion on adjusting the save behavior.

bwmilby (approved):

PR #464, opened 2026-08-25
Ensure unfolded state for script tabs

### LINKED ISSUE #463: folding persists in non-active script tabs if folding is turned off
When several script tabs are open and folding is turned off by the user the folded state persists in non-active tabs

BerndN:
ensure inactive script tabs unfold if folding is turned off
#463


emily-elizabeth (approved):
works as described

PR #467, opened 2026-08-25
removed erroneous text from the message box results

### LINKED ISSUE #465: The message box always starts with text in field "results"
When starting HXT 0.9.15 and earlier and one opens the message box it opens with
"dark,#323232,#ffffff"  in field "results".

It seems to be baked in.
A save of stack "Message Box" with an empty field "results" cures this on a local version of HXT.

emily-elizabeth:
removed erroneous text from the message box results


bwmilby (approved):
There are a bunch of additional changes in the binary stack other than removing the text but nothing looks "bad".  I think it is just a matter of the user's defaults getting baked into the file when it was saved.  This is why I'm working on that stripped down stack editor for FLC7, but it could work for OXT/HXT as well as long as careful about widgets.

PR #477, opened 2026-09-03
452 externals are not included in standalones

### LINKED ISSUE #452: externals are not included in standalones
when building a standalone, externals (revXML, revZIP, database drivers, etc) are not being included. in the HyperXTalk application bundle the "externals.txt" and "Database Drivers.txt" files are missing

emily-elizabeth:
Standalones built from the installed HyperXTalk app (outside the repo) had all their externals silently not load at runtime, even though the external files were correctly copied into the bundle

Three separate bugs combined to break external loading when building from an installed app (revEnvironmentIsInstalled() = true):

1. Tools/Externals/Externals.txt missing (primary cause)

revExternalsList() in installed mode reads Tools/Externals/Externals.txt to discover which externals are available. This file was never generated — the Makefile only generated Externals.txt in the Runtime folder. With the file missing, revExternalsList() returned empty, so the loop in revCopyExternals that copies externals and writes kMCCapsuleSectionTypeExternal capsule entries never ran. The built standalone had no external entries in its capsule and loaded nothing.

Fixed by adding a printf step to both make and package-mac-bin Makefile targets to generate Tools/Externals/Externals.txt, and creating the file directly in mac-bin for the current build.

2. Wrong revCopyModule arguments for revdb in installed mode (secondary cause)

The installed-mode block that force-copies revdb.bundle called revCopyModule with pStandaloneRoot = pDirectory & "/Externals" and pTargetFolder = "". This put the file in the right place but created the library mapping revdb:./revdb.bundle (no subfolder). At runtime the engine resolves that to MCappcodepath/revdb.bundle — which doesn't exist. Since the capsule processor aborts on the first failed loadexternal, this would have broken all subsequent externals too.

Fixed by changing the call to pStandaloneRoot = pDirectory, pTargetFolder = "Externals", producing the correct mapping revdb:./Externals/revdb.bundle.

3. ide/Runtime/Mac OS X/arm64/Externals/Externals.txt and Database Drivers.txt missing (prerequisite)

revExternalPath() and revDBDriverPath() in installed mode read from the Runtime folder to resolve the source paths of externals to copy. These files didn't exist in ide/Runtime/Mac OS X/arm64/Externals/, so source paths came back empty and all copies were skipped. Fixed in the previous commit by creating both files with content matching mac-bin.

### DIFF
--- a/Makefile.Mac
+++ b/Makefile.Mac
@@ -319,10 +319,8 @@ package-mac:
 	@# ----------------------------------------------------------------
 	@printf 'Speech,revspeech.bundle\nXML,revxml.bundle\nRevolution Zip,revzip.bundle\nDatabase,revdb.bundle\n' \
 	    > "$(TOOLS_DIR)/Externals/Externals.txt"
-	@printf 'Speech,revspeech.bundle\nXML,revxml.bundle\nRevolution Zip,revzip.bundle\n' \
+	@printf 'Speech,revspeech.bundle\nXML,revxml.bundle\nRevolution Zip,revzip.bundle\nDatabase,revdb.bundle\n' \
 	    > "$(RUNTIME_ARM64)/Externals/Externals.txt"
-	@printf 'Database,revdb.bundle\n' \
-	    >> "$(RUNTIME_ARM64)/Externals/Externals.txt"
 	@printf 'MySQL,dbmysql.bundle\nODBC,dbodbc.bundle\nPostgreSQL,dbpostgresql.bundle\nSQLite,dbsqlite.bundle\n' \
 	    > "$(RUNTIME_ARM64)/Externals/Database Drivers/Database Drivers.txt"
 	@# ----------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


bwmilby:
I don't think those .txt files should be committed to the repository.  They should be generated by script.  It is already there, just needs to add the other locations.
```
	@printf 'MySQL,dbmysql.bundle\nODBC,dbodbc.bundle\nPostgreSQL,dbpostgresql.bundle\nSQLite,dbsqlite.bundle\n' \
	    > "$(RUNTIME_ARM64)/Externals/Database Drivers/Database Drivers.txt"
```
Also, including the runtime files in the repository is just going to bloat it and make it unmanageable eventually.  Compiled code really should be stored elsewhere and downloaded to ignored locations in the repo.

emily-elizabeth:
I'm not seeing any runtime binary files being added. Did I miss something?

bwmilby:
Nothing in this PR, but there are Windows binaries in the repo.  Ideally we would be able to make a package for each runtime flavor that would be downloaded for creation of the IDE package.  Code somewhat exists to do it, but would take some rework to point it elsewhere.

bwmilby:
I think you still need to clone the db drivers file generation for the IDE copy.
I wonder why this was initially set up to do the copy like it did.  Seems like it would be easier to generate the entire Externals folder once and then just copy it from the IDE location to the standalone location.  Or could probably create a relative link from the standalone copy up to the IDE.  I know that I've seen links used in frameworks.

emily-elizabeth:
Seems like this issue affects Windows as well, will test on Linux after getting the Windows builds working

emily-elizabeth:
No it doesn't, just an issue with my interface that needs to be corrected.

emily-elizabeth:
Not an issue on Linux (Ubuntu)

bwmilby (approved):
Seems to work as expected.  When selecting externals now, they do make it into the Mac standalone app that is generated.

PR #479, opened 2026-09-03
Fix vanishing compilation errors

### LINKED ISSUE #478: Compile errors are not persistent if "LiveErrors" are off
Try this in an empty script and compile. It will throw an compilation error. When LiveErrors are off click anywhere and the compilation error will disappear. The error should be flagged regardless of the state of "LiveErrors" 

on foo
   put 5 then 
   end foo

BerndN:
fix vanishing compilation errors when "LiveError" is turned off

this fixes issue 
Compile errors are not persistent if "LiveErrors" are off
 #478


emily-elizabeth (approved):
Works as advertised

PR #485, opened 2026-09-05
481 get the height of the toolbar

### LINKED ISSUE #481: get the height of the toolbar
when using the toolbar, you need to resize your interface to remove the space taken up by the toolbar (Windows and Linux) so need a way to get the height of the toolbar as it will vary between platforms and distros

emily-elizabeth:
Problem

The Linux toolbar backend hardcoded kToolbarHeight = 48 in two places — the GdkWindow creation and the ConfigureNotify resize handler. This made toolbar height wrong across desktop environments (KDE ~38px, XFCE ~29px) and HiDPI displays (should be 96px at 2×). There was also no way for scripts to query the toolbar height to position card content below it, on any platform.

Solution

Add GetHeight() to the MCToolbarBackend interface and getToolbarHeight() (exposed as the toolbarHeight property) to MCToolbar, with correct implementations on all three platforms.

Linux — pre-creation, probes a throwaway GtkToolbar widget for its natural height from the active GTK theme, so DE-specific sizing is respected automatically. Post-creation, reads gdk_window_get_height() from the child window directly. Falls back to the compile-time constant only if GTK returns nothing. The two hardcoded kToolbarHeight uses in Create() and the resize handler are replaced.
Windows — GetWindowRect on the toolbar HWND after TB_AUTOSIZE, consistent with the logic already used in _reposition().
macOS — difference between the window frame rect height and the content rect height, which is what AppKit reserves for the unified NSToolbar.

Files changed

engine/src/toolbar.h — GetHeight() added to MCToolbarBackend; getToolbarHeight() declared on MCToolbar
engine/src/toolbar.cpp — getToolbarHeight() implemented
engine/src/lnx-toolbar.cpp — GetHeight() implemented; hardcoded kToolbarHeight sizing replaced
engine/src/w32-toolbar.cpp — GetHeight() implemented
engine/src/mac-toolbar.mm — GetHeight() implemented
docs/dictionary/property/toolbarHeight.lcdoc — new dictionary entry

Testing

On Linux, verify toolbar height is correct on GNOME, KDE, and XFCE, and on a HiDPI display.
Confirm the toolbarHeight of toolbar 1 returns a non-zero value when the toolbar is open, and 0 when the stack is closed.
Confirm card content positioned using toolbarHeight does not overlap the toolbar on any platform.


emily-elizabeth:
tested on macOS 26.6.2, Windows 11 and Ubuntu 25.04

emily-elizabeth:
tested on macOS 27 b8, macOS 14 and macOS 15; retested on Windows 11 and Ubuntu 25.04

emily-elizabeth:
tested on Linux Mint, Fedora, Debian (GNOME and MATE)

mwieder:
Hella warnings on launching, and the IDE is basically unusable because of the 'drawable' errors:

(HyperXTalk:1525045): GLib-GObject-CRITICAL **: 23:33:35.886: ../../../gobject/gsignal.c:2533: signal 'setting-changed' is invalid for instance '0x571a90bf1410' of type 'GdkX11Screen'

lots of instances of
(HyperXTalk:1525045): Gdk-WARNING **: 23:33:45.745: ../../../gdk/x11/gdkwindow-x11.c:5678 drawable is not a native X11 window

Note: this is in the main branch too, so not localized to this PR, but I just pulled several commits into the main branch and don't know where the messages are being generated.

emily-elizabeth:
Yeah, those were some missed from the web kit merge. It's been fixed in PR #500

mwieder:
Thanks. I'll wait for #500 integration then. Seems Brian's got it fixed up.

emily-elizabeth:
PR #500 has been merged to main

mwieder:
Still spewing all those errors/warnings though.

emily-elizabeth:
Yes, because this PR still uses the old main code. Did you want me to rebase it?

engine/src/exec-interface-object.cpp
```diff
@@ -759,6 +759,7 @@ static const PropList toolbarprops[] =
         {"lockLoc",          P_LOCK_LOCATION},
         {"name",             P_SHORT_NAME},
         {"rect",             P_RECTANGLE},
+        {"toolbarHeight",    P_TOOLBAR_HEIGHT},
```
mwieder: Why is this 'toolbarheight' and not just 'height'?

engine/src/exec-interface-object.cpp
```diff
@@ -759,6 +759,7 @@ static const PropList toolbarprops[] =
...
         {"toolbarVisible",   P_TOOLBAR_VISIBLE},
```
mwieder: likewise, why not just 'visible'?

mwieder:
I thought #500 was supposed to fix those

emily-elizabeth:
Let me check on the Linux test machine.

emily-elizabeth:
I can see the same warnings again. I'll make a new ticket as this needs to be resolved

PR #487, opened 2026-09-06
Feat toolbar shown hidden callbacks

emily-elizabeth:
Summary

Scripts can now respond to the toolbar being shown or hidden by handling two new messages: toolbarShown and toolbarHidden.

on toolbarShown
  set the height of card field "content" to the height of this card - the toolbarHeight
end toolbarShown

on toolbarHidden
  set the height of card field "content" to the height of this card
end toolbarHidden
Motivation

On Windows and Linux the toolbar lives inside the card's content area, so when it is shown or hidden the available layout space changes. Previously there was no way for a script to react to this automatically — handlers had to manually track the toolbar's visibility state. These messages close that gap.

Implementation
toolbarShown / toolbarHidden are dispatched from SetToolbarVisible through the current card, following the normal card → stack message path (same approach as toolbarItemClicked).
No message is sent when the visible property is set to its current value.
The messages fire during stack load, so the initial layout can reflect the saved toolbar state.
Files changed
engine/src/mcstring.h / mcstring.cpp — declared, defined, and registered the two new message names
engine/src/exec-interface-toolbar.cpp — added early-out and dispatch in SetToolbarVisible


emily-elizabeth:
Needs PR 485 to be merged first

emily-elizabeth:
tested on macOS, Windows and Ubuntu

PR #492, opened 2026-09-09
Update .gitignore

bwmilby:
Remove:
- prebuilt/include/zconf.h (file is tracked, does not make sense here)
- build-win-x86_64/* (covered earlier)
- mac-bin (covered earlier)
- build/engine/* (covered earlier)
- /thirdparty/libjpeg/src/jconfig.h (file tracked, no effect)

Untrack Linux compiled binary

moved a few lines into proper sections and split out exceptions from compiled source exclusions


emily-elizabeth (approved):
looks good. We'll have to discuss a plan to get the standalone runtimes to the other platforms so we can do compiling

bwmilby:
Agree on the standalone issue.  I think we should be able to phase it though.  Perform the compile step, "build" a standalone "package" (runtime folder contents) for that OS, upload somewhere.  Once all 3 compiles are done, go back and download the other 2 runtime packages, assemble the distribution app.  There is code that does this somewhat that we could probably leverage.

The file that I untracked was the IDE build which wouldn't be used on other OSes.

PR #496, opened 2026-09-10
495-lin-update-build-to-guard-against-gtk3-use-of-webp

### LINKED ISSUE #495: [lin] update build to guard against GTK3 use of WebP
Replace config/linux-settings.gypi line 28 and add a webp filter alongside the existing libpng one: The same line appears in engine/kernel.gyp:73, and kernel-server.gyp:154 with:
```
'<!@(pkg-config --cflags-only-I gtk+-3.0 2>/dev/null | sed "s/-I[^ ]*libpng[^ ]*//g" | sed "s/-I[^ ]*webp[^ ]*//g" | sed "s/-I//g")',
```

Replace engine/kernel.gyp:76 with:
```
'<!@(pkg-config --cflags-only-I gtk+-unix-print-3.0 2>/dev/null | sed "s/-I[^ ]*libpng[^ ]*//g" | sed "s/-I[^ ]*webp[^ ]*//g" | sed "s/-I//g")',
```

Regenerate
```
python3 config.py -DTARGET_PLATFORM=Linux -fmake
```

Verify before committing to another compile:
```
grep -c "unix-print" build-linux-x86_64/hyperxtalk/engine/kernel.target.mk
```
Non-zero means it took.

emily-elizabeth:
added sed to guard WebP


mwieder (commented):
Good timing. Just pulled in #495 and successfully finished a build.
I'm not altogether happy about this level of mucking with thirdparty libraries vs system libraries, but I can't think of a better way to do this.
I'd say this is good to go.

emily-elizabeth:
It was the way LiveCode had it for PNGs so we just included WebP

mwieder:
I realize LC had that in place since forever. Doesn't mean I have to like the mechanism. Just worry about painting ourselves into corners with almost-bespoke third-party libraries.

emily-elizabeth:
We can always look into a better way of doing it and update in the future 😄

PR #497, opened 2026-09-10
494-lin-meson-build-issue

### LINKED ISSUE #494: [lin] meson build issue
The fix in the script — make --wipe conditional:

```
if [ -e "build/build.ninja" ] ; then
    meson setup --wipe build
else
    meson setup build
fi
```

emily-elizabeth:
updated meson to build the first time it is run. On the first run, it tries to wipe the directory that doesn't exist so it fails.


emily-elizabeth:
I'll verify this with another clean build on Linux Mint

bwmilby:
I don't see the build/build.ninja directory.  Where would it be?  I've got Mint running, but compile stops due to something else (include not working correctly for webp).

emily-elizabeth:
For WebP, depending on when you cloned your repo, you may or may not have the fix - if not here are the instructions https://github.com/emily-elizabeth/HyperXTalk/issues/495

If the build/build.ninja directory doesn't exist it should run the setup and then it should have that directory.

bwmilby:
I don't know where to look though.  Should it be under the build directly under prebuilt?  I don't have it there and have a successful build.  Now, I did add `cmake` after an error which I guess could have been related.

emily-elizabeth:
/media/mint/repos/HyperXTalk/prebuilt/build/cairo-1.18.4/build/build.ninja

That is where mine is for Cairo, but you could have multiple folders if other libraries use it

emily-elizabeth:
Compiled successfully on Linux Mint with the meson fix in place. This was a clean build on a LiveBoot USB

bwmilby:
This is strange.  I just did a clean build on Linux Mint without this patch.  `build.ninja` is present in the path you mentioned.  All I did was an `apt-get update` and then the `apt install` from the other file I've been working on.  I used Cinnamon.  Before the patch, did make error out?

emily-elizabeth:
Yes, when I went to compile it caused an error when trying to wipe the directory, so I had to manually do the setup and then it compiled. The second time I did the compile, after rebooting and recloning the repo, I had the patch in place and it compiled successfully.

bwmilby:
I just did an interactive test.  For `--wipe` to be successful, the folder must be completely empty or it must be a prior build.  This check should be unnecessary for a completely fresh compile since the `mkdir` (line 110) will create the directory and it will be empty.  If there was stuff in there and you didn't delete everything, it will fail (I left a couple dot files and it failed).  The PR won't hurt anything though.

emily-elizabeth:
Not too certain what happened on my first compile that made it fail.

bwmilby:
> Yes, when I went to compile it caused an error when trying to wipe the directory, so I had to manually do the setup and then it compiled. The second time I did the compile, after rebooting and recloning the repo, I had the patch in place and it compiled successfully.

Really odd.  Mine was from a fresh clone, on main branch, no patch installed.
I had been trying to get a compile to work on a VM and it kept hanging when running server to compile LCB so I switched to using a different laptop and a thumb drive.

bwmilby (approved):
Compiles work as expected with the change.

PR #500, opened 2026-09-12
getting WebKitGTK to work on Fedora

### LINKED ISSUE #499: [lin] web browser not working on Fedora
Something happened between the last test and the merge to main, will need to look into this further

emily-elizabeth:
gtk_widget_draw() is a compositor no-op on WebKit 2.44+ (Fedora 40+), leaving doPaint() with only the white background fill.

Wire up the hxt-snap-fn/hxt-snap-ctx bridge:
- Load webkit_web_view_get_snapshot / webkit_web_view_get_snapshot_finish
- RequestSnapshot(): rate-limited async snapshot trigger (libbrowser)
- on_snapshot_done(): stores result as hxt-snapshot, calls hxt-repaint-fn
- TriggerRedraw(): schedules engine Redraw() when snapshot is ready (native-layer)
- doPaint(): detect blank surface and use cached snapshot as fallback

WebKit subprocesses inherit LD_LIBRARY_PATH from the AppImage runtime, loading AppImage-bundled libEGL instead of the system one. AppImage libEGL returns EGL_BAD_PARAMETER from eglGetDisplay(), aborting both WebKitWebProcess and WebKitGPUProcess.

Strip /tmp/.mount_ entries after our own dlopen() calls succeed.

linux: remove HXT- debug logging from libbrowser_webkitgtk


emily-elizabeth:
tested on Ubuntu, Linux Mint, Fedora, Debian GNOME and Debian MATE

emily-elizabeth:
Brian, do you have a Fedora VM or LiveBoot that you can test this on?

bwmilby:
I don't at the moment but just cleaned off a few 16GB thumb drives.  I should be able to set at least one of them up.  I'll try a live boot since I've had issues with compile on VM.

emily-elizabeth:
I'm using Fedora 22.3 Cinnamon so something along those lines

emily-elizabeth:
I have the ISO if you want me to put it on Mega

bwmilby:
I'm downloading the default version and Cinnamon.  Good thing is that the machine I'll use has a partition I should be able to mount to perform the compile.

emily-elizabeth:
Thanks a bunch :)

bwmilby:
I'm having a problem getting the compile due to an issue with g_free macro expansion.

bwmilby:
I had to add `#undef g_free` to `libbrowser_webkitgtk.cpp` to get `main` to compile on my Ubuntu 26.04.1 LTS system.  Was getting the same error on the Fedora machine (main or this branch).  Now it has an issue with libxml but starting a clean build now to make sure.  Took a bunch of tries to get everything installed to even try a compile (figuring out what was needed and names).

bwmilby:
It was an adventure, but finally did get a successful compile on Fedora.  Issue #506 initially, then `libxml2-devel` causing issues.  Edited a few `.mk` files to remove the header include.  I was able to launch and instantiate a working browser.  Still see a good bit of debug code being printed.  There was a GDK-warning about `drawable is not a native X11 window` in `gdk/x11/gdkwindow-x11.c:5688` quite a number of times.  Not sure if it is related to webkit since it was the first compile on that OS.  I would need to back down to the WebP commit and compile to figure that out.

On the same branch on my Ubuntu seat, I still have to make the `g_free` edit, but otherwise compiles fine.  I do not see the same level of debug and do not see that GDK-warning.

emily-elizabeth:
All the warnings have been suppressed

bwmilby:
Agree, all warning text has been suppressed.  Only code change I had to make to get it to compile was the `#undef g_free` that I mentioned earlier.  I was able to get a browser window to work in the IDE on Fedora.

bwmilby (approved):

PR #515, opened 2026-09-18
503 lin script editormessage box slow typing

### LINKED ISSUE #503: [lin] script editor/message box, slow typing
when typing in the script editor or the message box, it's really slow

emily-elizabeth:
Two root causes were identified via GDB captures during the freeze window:

1. GTK font cache (engine/src/linux-theme.cpp)
   The Script Editor's syntax highlighter called textHeight in a tight
   loop, triggering gtk_style_context_get() on every iteration. Under
   WebKitGTK, WebKit's CSS providers make this call ~100x slower than
   under CEF. Fixed by caching PangoFontDescription* per control type,
   invalidated via the existing MCLinuxThemeFlushCache().

2. X11 clipboard timeout (engine/src/lnx-clipboard.cpp, clipboard.cpp)
   On every keystroke the Script Editor issues a `select` command, which
   caused MCClipboard::AddText() to call PullUpdates() before writing.
   PullUpdates() sends XConvertSelection to the current PRIMARY owner and
   waits up to 1000ms for SelectionNotify. On XWayland, Wayland-native
   apps holding PRIMARY never respond to X11 CONVERT_SELECTION requests,
   so the full timeout was hit every keystroke.
   - Reduced WaitForSelectionNotify() timeout from 1000ms to 150ms
   - Added p_skip_pull to AutoLock so AddText() (a write path) skips
     PullUpdates() entirely — it's about to replace the clipboard anyway

Also fixed a keyDown deferral race condition in the Script Editor
(revsecommoneditorbehavior.livecodescript): pChar is now baked into the
`send` message string so rapid keystrokes each carry their own character
rather than sharing a script local that could be overwritten before the
deferred handler fires.

Diagnostic/debug code removed:
- [HXT-TRACE] backtrace blocks and execinfo.h in libbrowser_webkitgtk.cpp
- Diagnostic `return` in MCLinuxWebKitGTKRunloopAction (re-enables WebKit
  event loop — confirmed not the cause of the lag)
- [HXT-TIMING] instrumentation in lnxstack.cpp
- Timing `put ... ms after msg` probes in revsecommoneditorbehavior.livecodescript
- libhxt_trace.c (no longer referenced)


mwieder:
I dunno. Maybe I resolved the conflicts wrong here.
It's better but not solved.
And copy/paste no longer work.

In particular, copy seems to work but paste does not. I can copy text from a script editor and paste it into a text file but I can't paste it back into the script editor or into the message box.

emily-elizabeth:
> In particular, copy seems to work but paste does not. I can copy text from a script editor and paste it into a text file but I can't paste it back into the script editor or into the message box.

yeah, I noticed that too. If you copy and paste into Text Edit and then copy from Text Edit you can paste into the script editor (you need to do that second copy though) We're looking into it.

emily-elizabeth:
tested on Ubuntu, Fedora, Linux Mint, Debian GNOME and Debian MATE.
typing speed seemed responsive and was able to keep up with my typing (which isn't the fastest in the world, but it'll do in a pinch)
cut, copy and paste worked with controls on the stack window, script editor and message box

bwmilby (approved):
I cannot comment on the code itself as it would take me far to long to understand all of the changes in the C++ source.  Comments seem reasonable.  I was able to compile on my Ubuntu iMac and the speedup in text handling was noticeable.  I verified that copy/paste worked to/from script editor.

mwieder:
Ditto on commenting on most of the code. But glancing at it, it appears that the messagebox fix in revmessageboxsinglelinemessagebehavior.livecodescript is unrelated to the script editor fix, yes?

emily-elizabeth:
At one point I thought maybe the auto-complete was causing the slow down so that's what the change is about.

PR #520, opened 2026-09-20
fix: return 0 from MCWindowProc after WM_CLOSE to prevent DefWindowProc destroying IDE windows

### LINKED ISSUE #519: [win] unable to open Settings a second time
- open Edit > Settings
- close the window
- open Edit > Settings (no window)

emily-elizabeth:
The global hotkeys refactor in 0.9.15 (c33599a5c) restructured MCWindowProc so that all switch cases fall through to a single DefWindowProc call at the end of the function. Previously, explicitly handled messages returned 0; after the refactor they all call DefWindowProc.

For WM_CLOSE this is a silent regression: DefWindowProc's default behaviour for WM_CLOSE is to call DestroyWindow, which destroys the OS-level window unconditionally — even when the engine's own wclose() handler has already processed the message and the script cancelled the close via closeStackRequest. The result is that IDE palette/modeless windows (settings, property inspector, script editor, etc.) are destroyed at the OS level on first close and cannot be reopened, with no error or visible indication.

Fix: add an explicit `return 0` to the WM_CLOSE case so that after wclose() has dispatched closeStackRequest and applied the script's decision (close or cancel), Windows receives no further instruction. wclose() already contains the complete close-or-cancel decision tree including calling target->close() and target->checkdestroy() when the script passes; there is nothing for DefWindowProc to add.

Claude-Session: https://claude.ai/code/session_01JnjnNvMARU5z6YwW1PbTDk


emily-elizabeth:
tested on Windows 11
global hot keys work as well as closing and reopening IDE windows

PR #522, opened 2026-09-21
Fix deferred keyDown dropping quote and special characters

### LINKED ISSUE #521: [lin] unable to type doube-quote in script editor
- open a script editor
- type > answer "moo"
- the quotes do not appear

emily-elizabeth:
The keyDown deferral was embedding pChar in the send string wrapped in quote delimiters. When pChar itself was a double-quote character, the resulting string (""") would be misparse by LiveCode's message parser, causing the character to arrive empty in the handler and be silently dropped. The same mechanism could affect other characters that interact with LiveCode's string parsing.

Fix: pass pChar as its Unicode code point via charToNum() on the send side and reconstruct with numToChar() in __DoKeyDown_Deferred. Since the argument is now a plain integer, no quoting is needed and the full Unicode range — including ", (, ), emoji, and accented characters — round-trips cleanly.


emily-elizabeth:
This one does not need to be re-compiled as it's just a change to a behaviour script.

emily-elizabeth:
it affects all three platforms, but the fix works on all three
tested on macOS 26.6.2, Windows 11 and Ubuntu 25.04

mwieder (approved):
OK. Good by me.
I was thinking that numToByte and byteToNum would be faster, but for keystrokes the few milliseconds saved won't make a difference.

PR #526, opened 2026-09-22
We no longer need or want to build the cef library

mwieder:
I removed the version files since we are no longer building the cef library.


bwmilby:
Looks good

